PR #99, opened 2017-12-07
♲ Renamed properties and added additional constraints

Shodske:


specification/includes/email.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "string",
+  "example": "john@doe.com",
+  "pattern": "^(([^<>()\\[\\]\\.,;:\\s@\"]+(\\.[^<>()\\[\\]\\.,;:\\s@\"]+)*)|(\".+\"))@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$"
```
Shodske: I got this pattern from http://emailregex.com/
This should work for almost all emails, but if we want a less strict pattern, we could also change it to something like:
`"^(.)+@(.)+\.(.)+$"`

specification/info.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.2.27",
+  "version": "0.3.0",
```
Shodske: Changing it to `0.3.0`, because its breaking backwards compatibility.

specification/properties/SystemMessage.json
```diff
@@ -9,9 +9,13 @@
     ],
     "example": "info"
   },
-  "filter": {
+  "resource_type": {
```
Shodske: This is the resource type that this message is related to, which could indicate to the backoffice where to display this message.

specification/definitions/Webhook.json
```diff
@@ -15,14 +15,19 @@
...
       "required": [
+        "code",
         "name",
         "description"
       ],
       "additionalProperties": false,
       "properties": {
+        "code": {
+          "type": "string",
+          "example": "shipment_status_update"
```
Shodske: I created a `code` property here to identify the webhook, now we can keep the name and description more readable for humans and we can translate the name and description, but keep the `code` the same.

specification/includes/email.json
nessy: (.)+(.)+(.) = 3 8008's

specification/definitions/Webhook.json
nessy: This also had to change in the API. Is there a Utrecht item for this?

specification/definitions/Status.json
```diff
@@ -36,10 +36,9 @@
           ],
           "example": "shipment_delivered"
         },
-        "resource": {
+        "resource_type": {
```
nessy: Did you already change this in API as well?

specification/definitions/File.json
```diff
@@ -15,12 +15,12 @@
...
       "type": "object",
       "required": [
-        "resource_type",
+        "document_type",
         "formats"
       ],
       "additionalProperties": false,
       "properties": {
-        "resource_type": {
+        "document_type": {
```
nessy: Did you already change this in API as well?

specification/includes/email.json
nessy: No 8008's ??

PR #124, opened 2018-01-31
MP-171 ♻️ altered descriptions and made data/attributes required

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-171
plz rlz 0.5.2

### DIFF
--- a/specification/definitions/Service.json
+++ b/specification/definitions/Service.json
@@ -55,11 +55,11 @@
           "additionalProperties": false,
           "properties": {
             "min": {
-              "type": ["number", "null"],
+              "type": "number",
               "example": 1
             },
             "max": {
-              "type": ["number", "null"],
+              "type": "number",
               "example": 3
             }
           }
--- a/specification/definitions/Service.json
+++ b/specification/definitions/Service.json
@@ -18,7 +18,6 @@
       "required": [
         "name",
         "package_type",
-        "transit_time",
         "handover_method"
       ],
       "additionalProperties": false,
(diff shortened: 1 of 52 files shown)


specification/includes/physical-properties.json
```diff
@@ -27,7 +27,7 @@
       "type": "number",
       "minimum": 0,
       "example": 22.5,
-      "description": "Volume in litres"
+      "description": "Volume in litres (dm3)"
```
Shodske: liters*

specification/includes/read-only-physical-properties.json
```diff
@@ -25,7 +25,7 @@
       "type": "number",
       "minimum": 0,
       "example": 22.5,
-      "description": "Volume in litres"
+      "description": "Volume in litres (dm3)"
```
Shodske: liters*

specification/info.json
```diff
@@ -1,7 +1,7 @@
 {
-  "version": "0.4.8",
+  "version": "0.5.1",
```
Shodske: `0.5.2` when this is released?

specification/definitions/Service.json
```diff
@@ -55,11 +55,11 @@
           "additionalProperties": false,
           "properties": {
             "min": {
-              "type": ["number", "null"],
+              "type": "number",
```
Shodske: Not a comment for this line, but `transit_time` is currently a required property, this should not be required (we don't send`null` values).

PR #144, opened 2018-02-27
:recycle: updated regex from region code to also include empty string

nessy:
Is needed for MP-294


specification/schemas/RegionCode.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "string",
-  "pattern": "^[A-Z]{2,3}$",
+  "pattern": "^$|^[A-Z]{2,3}$",
```
Shodske: See my other comment.

PR #147, opened 2018-03-02
MP-325 Add combined files to api spec

NickVries:
[MP-325](https://myparcelcombv.atlassian.net/browse/MP-325)

Part of: [MP-292](https://myparcelcombv.atlassian.net/browse/MP-325)

### DIFF
--- a/specification/schemas/CombinedFiles.json
+++ b/specification/schemas/CombinedFiles.json
@@ -1,0 +1,85 @@
+{
+  "type": "object",
+  "allOf": [
+    {
+      "$ref": "#/components/schemas/MinimalCombinedFiles"
+    },
+    {
+      "required": [
+        "attributes",
+        "relationships"
+      ],
+      "properties": {
+        "attributes": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "document_type": {
+              "type": "string",
+              "enum": [
+                "label"
+              ],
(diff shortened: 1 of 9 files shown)


specification/schemas/CombinedFiles.json
```diff
@@ -0,0 +1,85 @@
...
+            "starting_position": {
+              "type": "string",
+              "enum": [
+                "top-left",
+                "top-right",
+                "bottom-left",
+                "bottom-right"
+              ],
+              "example": "top-left"
+            }
```
Shodske: You can add `"default": "label"`, `"default": "A4"` and `"default": "top-left"` to these attributes to indicate what the default values will be when not supplied.

specification/schemas/CombinedFiles.json
```diff
@@ -0,0 +1,85 @@
...
+                },
+                {
+                  "readOnly": true
+                }
+              ]
+            },
+            "source_files": {
+              "type": "array",
+              "items": {
+                "$ref": "#/components/schemas/FileRelationship"
```
Shodske: The way you've defined the `FileRelationship` schema means that this array wil look like this:
```
"source_files": [
  {"data": {"type": "files", "id": "1"}},
  {"data": {"type": "files", "id": "2"}},
  {"data": {"type": "files", "id": "3"}}
]
```

PR #154, opened 2018-03-14
MP-337 :sparkles: Added shipment patch endpoint

nessy:


specification/paths/Shipments-shipment_id.json
```diff
@@ -49,6 +49,59 @@
...
+        "application/vnd.api+json": {
+          "schema": {
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "$ref": "#/components/schemas/Shipment"
```
Shodske: Won't this require the user to always supply the entire shipment when patching?

M4tini: Yup, this is the case for all `PATCH` endpoints, probably introduced with the OpenApi migration. I created task MP-621 to fix this implementation.

PR #163, opened 2018-03-29
MP-623 :sparkles: added hooks endpoints

nessy:


specification/schemas/Action.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "object",
+  "required": [
+    "action_type",
+    "values"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "action_type": {
+      "type": "string",
+      "example": "update-query"
```
Shodske: `update-query` was removed from the proposal and is not something we will (initially) support.

specification/schemas/PredicateValue.json
```diff
@@ -0,0 +1,32 @@
+{
```
Shodske: Should this schema just be called `Value`? As it's not just a value for a predicate, but a more generic value, which in turn can be a predicate.

specification/schemas/PredicateValue.json
```diff
@@ -0,0 +1,32 @@
...
+       "enum": [
+         "path",
+         "scalar",
+         "predicate",
+         "resource"
+       ],
+       "example": "scalar"
+    },
+    "value": {
+      "type": "object",
```
Shodske: The value is not always an object, when the `value_type` is `scalar`, this can be a `number`, `string`, or `null`. And when the `value_type` is `path`, this is a `string`.

specification/schemas/PredicateValue.json
Shodske: Maybe it's a good idea to split up the different value types into `PredicateValue`, `PathValue`, `ScalarValue` and `ResourceValue` and have the generic `Value` just be:
```
{
  "oneOf": [
    {"$ref": "./ScalarValue"},
    ...
  ]
}

File: specification/schemas/Action.json [hunk 4]
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "resource_type": {
+      "type": "string",
+      "example": "services"
+    },
+    "values": {
+      "type": "array",
+      "items": {
+        "$ref": "#/components/schemas/PredicateValue"
```
Shodske: In my proposal these weren't just values, but (I called the database table) `path_values`. Because they didn't just represent values, but values assigned to a specific path in the context of the resource.

specification/schemas/Hook.json
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  "properties": {
+    "resource_type": {
+      "type": "string",
+      "example": "services"
+    },
+    "resource_action": {
+      "type": "string",
+      "enum": [
+        "create","read","update","delete"
```
M4tini: Put enum values on their own row.

specification/schemas/Predicate.json
```diff
@@ -0,0 +1,24 @@
...
+    "operator",
+    "left_value",
+    "right_value"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "operator": {
+      "type": "string",
+      "enum": [
+        "==", "!=", ">", "<", ">=", "<=", "AND", "OR"
```
M4tini: Put enum values on their own row.

specification/schemas.json
```diff
@@ -119,6 +125,12 @@
   "PickupDropoffLocation": {
     "$ref": "./schemas/PickupDropoffLocation.json"
   },
+  "Predicate": {
```
M4tini: Should be placed after `PhysicalProperties` to keep the list ordered alphabetically.

M4tini:
I added a commit to fix the stuff I found.

specification/schemas/Action.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "object",
+  "required": [
+    "action_type",
+    "values"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "action_type": {
+      "type": "string",
+      "example": "create"
```
Shodske: I think we should make this an `enum` with `create-resource` and `update-resource`, to make it clear that these are the 2 actions we (initially) support.

specification/schemas/Hook.json
```diff
@@ -0,0 +1,31 @@
...
+      "type": "string",
+      "example": "services"
+    },
+    "resource_action": {
+      "type": "string",
+      "enum": [
+        "create",
+        "read",
+        "update",
+        "delete"
```
Shodske: We're only going to support `create` and `update` initially. So for now I think it's best to remove the `delete` and `update` resource actions, to keep people from trying to create hooks that don't work (yet).

specification/schemas/Predicate.json
```diff
@@ -0,0 +1,57 @@
...
+        {
+          "$ref": "#/components/schemas/ScalarValue"
+        },
+        {
+          "$ref": "#/components/schemas/PathValue"
+        },
+        {
+          "$ref": "#/components/schemas/ResourceValue"
+        }
+      ]
```
Shodske: Instead of doing this `oneOf` everywhere, you could create a `Value` schema that holds this `oneOf` and reference that everywhere.

specification/schemas/ResourceValue.json
```diff
@@ -0,0 +1,32 @@
...
+        "resource_type": {
+          "type": "string",
+          "example": "regions"
+        },
+        "path": {
+          "type": "string",
+          "example": "relationships.parent.relationships.parent"
+        },
+        "attributes": {
+          "type": "object",
```
Shodske: The attributes in the proposal have dynamic properties. Now that I think about it, that may not be the best way as now you cannot define how these attributes should look, so this may need to be changed from the proposal.

specification/schemas/Action.json
```diff
@@ -0,0 +1,48 @@
...
+          },
+          {
+            "$ref": "#/components/schemas/ScalarValue"
+          },
+          {
+            "$ref": "#/components/schemas/PathValue"
+          },
+          {
+            "$ref": "#/components/schemas/ResourceValue"
+          }
```
Shodske: If you look at the proposal, the values here aren't just values, but have a `path` and then a `value` to indicate what value should be assigned to what value.

specification/schemas/ResourceValue.json
nessy: What did you have in mind? adding an oneOf and add all the resource types?

specification/schemas/Hook.json
```diff
@@ -0,0 +1,29 @@
+{
+  "type": "object",
+  "required": [
+    "resource_type",
+    "resource_action",
+    "predicate",
+    "action"
```
Shodske: I just realized, a Hook should of course be a proper resource, with an `id`, `type`, `attributes` and `relationships`. And I think it should actually have a relationship with either a `shops`, `organizations`, `brokers` or `enterprises` resource. Although I can't think of a good name for the relationship off the top of my head.

nessy: And the organizations, brokers and enterprises do not exits jet.

specification/schemas/MinimalEnterprise.json
```diff
@@ -0,0 +1,19 @@
+{
+  "type": "object",
+  "required": [
+    "type"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
+      "pattern": "^enterprise$",
```
Shodske: This should be plural `enterprises`

specification/schemas/MinimalOrganization.json
```diff
@@ -0,0 +1,19 @@
+{
+  "type": "object",
+  "required": [
+    "type"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
+      "pattern": "^organization$",
```
Shodske: `organizations`

specification/schemas/Owner.json
```diff
@@ -0,0 +1,16 @@
+{
```
Shodske: Same as `MinimalOwner`, there is no owner type, so this should be removed.

specification/schemas/Broker.json
```diff
@@ -0,0 +1,45 @@
+{
```
Shodske: I think we should omit the `Broker`, `Organization` and `Enterprise` schemas, until we have endpoints that actually return those resources.

specification/schemas/MinimalOwner.json
```diff
@@ -0,0 +1,16 @@
+{
```
Shodske: There is no owner type, so there is no 'minimal owner', there is only a relationship with the name `owner`. So this should be removed.

specification/schemas/ScalarValueType.json
```diff
@@ -0,0 +1,15 @@
...
+  "required": [
+    "value_type",
+    "value"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "value_type": "scalar",
+    "value": {
+      "type": "number",
+      "example": "5000"
```
Shodske: I scalar value is not just a number, but can be a number, string or null.

PR #164, opened 2018-04-04
MP-626 Customs invoice number requirements

NickVries:
[MP-626](https://myparcelcombv.atlassian.net/browse/MP-626)

Part of: [MP-594](https://myparcelcombv.atlassian.net/browse/MP-594)

### DIFF
--- a/specification/schemas/InvoiceCustoms.json
+++ b/specification/schemas/InvoiceCustoms.json
@@ -1,0 +1,43 @@
+{
+  "type": "object",
+  "required": [
+    "content_type",
+    "invoice_number",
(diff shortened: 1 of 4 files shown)


specification/schemas/InvoiceCustoms.json
```diff
@@ -0,0 +1,43 @@
+{
+  "type": "object",
```
SAMDevelopment: You can make this schema extend the previous schema. That way we won't have to maintain the overlapping fields in two places.

[DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

PR #165, opened 2018-04-05
:bug: Fixed bug with enum values on invoice custom shipments.

NickVries:


specification/schemas/BaseCustoms.json
```diff
@@ -5,16 +5,7 @@
     "non_delivery",
     "incoterm"
   ],
-  "additionalProperties": false,
```
Shodske: You can probably keep ` "additionalProperties": false` in here if you add `content_type` as an enum with all the possibilities.

PR #169, opened 2018-04-11
MP-758 Updated hooks and removed them from the path

nessy:


Shodske:
Closing for now as it is still work in progress.

specification/paths/Webhooks-webhook_id.json
```diff
@@ -1,7 +1,7 @@
 {
   "get": {
     "tags": [
-      "Webhooks"
+      "⚠ Webhooks"
```
nessy: Should hooks not be removed completely?

specification/paths/Webhooks.json
```diff
@@ -1,7 +1,7 @@
 {
   "get": {
     "tags": [
-      "Webhooks"
+      "⚠ Webhooks"
```
nessy: Should hooks not be removed completely?

specification/paths/Webhooks-webhook_id.json
Shodske: Removing them completely right now will cause the API tests to fail.

PR #177, opened 2018-04-30
:sparkles: MP-778 Add billing attributes to organization schema

denniskoster:


specification/schemas/Organization.json
```diff
@@ -23,6 +23,37 @@
         "iban": {
           "$ref": "#/components/schemas/Iban"
         },
+        "language": {
+          "type": "enum",
+          "enum": [
+            "english"
+          ],
+          "example": "english"
```
Shodske: @SAMDevelopment @nessy @M4tini @NickVries @denniskoster 
Did we decide on how we wanted to have languages in the API? I think it's better to use ISO codes here instead of the names of the language (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). Mostly because we're also using ISO codes for currencies and countries etc.

NickVries: Besides, English is with a capital E.

denniskoster: I agree with @Shodske. Maybe it even makes sense to create a separate language resource for future association with countries?

M4tini: I agree with ISO_639-1 two-letter codes. And I think we don't need a language resource + endpoint + relationships, because this will add complexity and the requirement to use our language_id instead of the ISO code.

specification/schemas/Organization.json
```diff
@@ -23,6 +23,37 @@
...
+          ],
+          "example": "prepayment"
+        },
+        "vat_number": {
+          "type": "string",
+          "example": "123456789MVA"
+        },
+        "payment_term": {
+          "type": "integer",
+          "example": "1"
```
M4tini: It would be nice to have a description here. Does this `1` mean a day or a month?

specification/schemas/Organization.json
```diff
@@ -23,6 +23,37 @@
         "iban": {
           "$ref": "#/components/schemas/Iban"
         },
+        "language": {
+          "type": "enum",
```
Shodske: In json schema, `enum` is not a type, but any type can be an enum. In this case the type should be `string`.

specification/schemas/Organization.json
```diff
@@ -23,6 +23,37 @@
...
+          "enum": [
+            "english"
+          ],
+          "example": "english"
+        },
+        "payment_method": {
+          "type": "enum",
```
Shodske: `"type": "string"`

specification/schemas/Organization.json
```diff
@@ -23,6 +23,37 @@
...
+            "prepayment"
+          ],
+          "example": "prepayment"
+        },
+        "vat_number": {
+          "type": "string",
+          "example": "123456789MVA"
+        },
+        "payment_term": {
+          "type": "integer",
```
Shodske: There is no integer (or at least not officially) in json schema and this should be `number`.

Shodske:
You need to rebase this branch on master.

specification/schemas/Organization.json
```diff
@@ -27,6 +27,30 @@
...
+              "example": "prepayment"
+            },
+            "vat_number": {
+              "type": "string",
+              "example": "123456789MVA"
+            },
+            "payment_term": {
+              "type": "number",
+              "example": "1",
+              "description": "Amount of days in which organization needs to pay"
```
NickVries: an/the organization*

PR #186, opened 2018-05-04
Add broker relationship to organization schema.

NickVries:
Part of: [MP-639](https://myparcelcombv.atlassian.net/browse/MP-639)

### DIFF
--- a/specification/paths/Organizations-organization_id.json
+++ b/specification/paths/Organizations-organization_id.json
@@ -23,7 +23,32 @@
               "additionalProperties": false,
               "properties": {
                 "data": {
-                  "$ref": "#/components/schemas/Organization"
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/Organization"
+                    },
+                    {
+                      "required": [
+                        "attributes",
+                        "relationships"
+                      ],
+                      "properties": {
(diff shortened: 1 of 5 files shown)


specification/paths/Organizations-organization_id.json
```diff
@@ -23,7 +23,32 @@
...
+                      "$ref": "#/components/schemas/Organization"
+                    },
+                    {
+                      "required": [
+                        "attributes",
+                        "relationships"
+                      ],
+                      "properties": {
+                        "id": {
+                          "readOnly": true
```
Shodske: The `id` in the `get` and `patch` endpoints should be required. I don't know if they are already required, but in swagger ui it looks like they are not.

specification/paths/Organizations-organization_id.json
```diff
@@ -23,7 +23,32 @@
...
                 "data": {
-                  "$ref": "#/components/schemas/Organization"
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/Organization"
+                    },
+                    {
+                      "required": [
+                        "attributes",
+                        "relationships"
```
Shodske: `links` and `links.self` should be required in the get endpoints.

PR #187, opened 2018-05-07
:bug: carrier contract enum added missing example

nessy:


specification/schemas/CarrierContract.json
```diff
@@ -17,15 +17,21 @@
               "$ref": "#/components/schemas/Currency"
             },
             "active": {
-              "type": "boolean"
+              "type": "boolean",
+              "example": true
+            },
+            "name": {
+              "type": "string",
+              "example": "Tier 1"
```
Shodske: Lets not use "Tier 1" as an example, but a name someone may give their own contract, like "My DPD contract".

PR #196, opened 2018-05-21
Add nett_weight attribute to shipment item resource

denniskoster:


specification/schemas/ShipmentItem.json
```diff
@@ -2,7 +2,8 @@
   "type": "object",
   "required": [
     "description",
-    "quantity"
+    "quantity",
+    "nett_weight"
```
Shodske: I don't think it should always be required, but it should only be required in an international shipment.

denniskoster: Aren't shipments with shipment items always international shipments? At least from a backoffice point of view it is I think. You only get the shipment items form when selecting a foreign country to ship to.

Shodske: The API allows you to add items to any shipment, they are just not required. This allows customers to keep track of what was in what shipment. The back-office doesn't give you the option to add items because they are not required for regular shipments, and we don't think users of the back-office will want to manually add items to a shipment if they don't have to.

PR #197, opened 2018-05-23
MP-621 Fix patch endpoints

NickVries:
[MP-621](https://myparcelcombv.atlassian.net/browse/MP-621)

Api task is also done, but waiting on this PR.

After merging: 
- [ ] Make new release.


specification/schemas/BaseShipment.json
```diff
@@ -0,0 +1,197 @@
...
+            },
+            "register_at": {
+              "oneOf": [
+                {
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/Timestamp"
+                    },
+                    {
+                      "readOnly": false,
```
Shodske: Why is there a `"readOnly": false` here?

specification/schemas/BaseAddress.json
```diff
@@ -0,0 +1,86 @@
...
+      "type": [
+        "string",
+        "null"
+      ],
+      "example": "Holmes"
+    },
+    "company": {
+      "type": [
+        "string",
+        "null"
```
Shodske: In other parts you're using `"oneOf"` instead of a type array, is there a reason that it's different here?

specification/schemas/Shipment.json
```diff
@@ -10,8 +10,10 @@
         "relationships"
       ],
       "properties": {
+        "id": {
+          "readOnly": true
+        },
         "attributes": {
-          "type": "object",
```
Shodske: Should this be removed?

specification/schemas/BaseAddress.json
NickVries: In other parts some cases where I use a `"oneOf"`, it's because that attribute has enum values. In order for null to be accepted as a value in an enum, you need to add it to the enum. In the UI however, that would show up as: `"enum": [ first, second, ].` Notice the comma with empty space behind it in the array. That's why I chose for oneOf there. In general, I prefer to use the type array for nullable values, so that's why I did it differently here, because it's not an issue here. It is a bit inconsistent. I agree.

specification/schemas/BaseShipment.json
NickVries: The Timestamp schema has `"readOnly": true` inside it, because of the `created_at` and `updated_at` timestamp values being `readOnly`. I added `"readOnly": false` here, instead of adding `"readOnly": true` to the other times the `Timestamp` schema is used because it's more common for a `Timestamp` to be `readOnly` than not.

specification/schemas/Shipment.json
NickVries: Nope, that was leftover from when I tried to extend the `BaseShipment` schema. I'll add it back.

specification/schemas/Shipment.json
```diff
@@ -131,7 +89,24 @@
                   "$ref": "#/components/schemas/PhysicalProperties"
                 },
                 {
-                  "readOnly": true
+                  "readOnly": true,
```
nessy: Can properties have readonly?

specification/schemas/BaseShipment.json
Shodske: But won't this break when I try to post a shipment with `register_at` set? I don't know if readonly is enforced anywhere, but if the shipment is to comply with **all*Of these schema's it will fail at the `register_at` not being readonly.

NickVries: readOnly is not enforced. I just tried it, and even with readOnly true it still lets it through... In the ui though, it shows readOnly false instead of true.

PR #199, opened 2018-05-23
Added contract post, delete and patch endpoint

nessy:
[MP-440](https://myparcelcombv.atlassian.net/browse/MP-440)

### DIFF
--- a/specification/examples/HookRequest.json
+++ b/specification/examples/HookRequest.json
@@ -1,7 +1,6 @@
 {
   "type": "hooks",
   "attributes": {
-    "type": "object",
     "name": "on-shipment-create-update-description",
     "resource_type": "shipments",
     "resource_action": "create",
--- a/specification/examples/HookRequest.json
+++ b/specification/examples/HookRequest.json
@@ -1,6 +1,7 @@
 {
   "type": "hooks",
   "attributes": {
+    "type": "object",
     "name": "on-shipment-create-update-description",
     "resource_type": "shipments",
     "resource_action": "create",
--- a/specification/examples/PostShipmentRequest.json
+++ b/specification/examples/PostShipmentRequest.json
@@ -1,6 +1,7 @@
 {
   "type": "shipments",
   "attributes": {
+    "type": "object",
     "recipient_address": {
       "street_1": "Baker Street",
       "street_2": "Marylebone",
--- a/specification/examples/PostShipmentRequest.json
+++ b/specification/examples/PostShipmentRequest.json
@@ -1,7 +1,6 @@
 {
   "type": "shipments",
   "attributes": {
-    "type": "object",
     "recipient_address": {
       "street_1": "Baker Street",
       "street_2": "Marylebone",
--- a/specification/examples/ShipmentResponse.json
+++ b/specification/examples/ShipmentResponse.json
@@ -6,7 +6,6 @@
     {
       "id": "7b808eee-bf1c-40cd-98f2-3c335a06417e",
       "attributes": {
-        "type": "object",
         "price": {
           "amount": 995,
           "currency": "EUR"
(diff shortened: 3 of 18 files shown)


specification/paths/CarrierContracts-carrier_contract_id.json
```diff
@@ -39,5 +39,97 @@
...
+                  {
+                    "required": [
+                      "id"
+                    ],
+                    "properties": {
+                      "attributes": {
+                        "properties": {
+                          "credentials": {
+                            "type": "string",
+                            "example": "{\"api_user\":\"your_api_user\",\"api_password\":\"your_password\"}"
```
NickVries: This example looks weird. It seems that escaping double quotes isn't possible in our spec. Have you considered just using type "object" and then in the API converting the object into json string? That also looks nicer for the person making the request. Instead of writing one long string of JSON.

specification/paths/CarrierContracts-carrier_contract_id.json
```diff
@@ -39,5 +39,97 @@
...
+          }
+        }
+      }
+    }
+  },
+  "delete": {
+    "tags": [
+      "CarrierContracts"
+    ],
+    "summary": "Delete an carrier contract by id.",
```
NickVries: Here, it's **a** carrier contract:  Delete a carrier contract by id.

specification/paths/CarrierContracts-carrier_contract_id.json
```diff
@@ -39,5 +39,97 @@
...
+          "application/vnd.api+json": {
+            "schema": {
+              "type": "object",
+              "required": [
+                "data"
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "data": {
+                  "$ref": "#/components/schemas/CarrierContract"
```
NickVries: For this, and other response bodies (in get all, get by id and post) the attributes: currency, name, status and active are specified as `not null` in the database. So you will always receive some value for these in a response. Therefore, they can be marked as required in the response body. You can achieve this with an "allOf".

specification/paths/CarrierContracts.json
```diff
@@ -0,0 +1,148 @@
...
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "allOf": [
+                  {
+                    "$ref": "#/components/schemas/CarrierContract"
+                  },
+                  {
+                    "properties": {
```
NickVries: Since the database specifies that the currency and name columns cannot be null, they should be required within the  body of a post request. (the active column also cannot be null, but has a default value `false`). You can add that here in the `allOf`.

specification/paths/CarrierContracts-carrier_contract_id.json
nessy: The status attribute is read only so that one we set internally in the Api

specification/examples/HookRequest.json
```diff
@@ -1,6 +1,7 @@
 {
   "type": "hooks",
   "attributes": {
+    "type": "object",
```
NickVries: This should not be here.

specification/examples/PostShipmentRequest.json
```diff
@@ -1,6 +1,7 @@
 {
   "type": "shipments",
   "attributes": {
+    "type": "object",
```
NickVries: This should not be here.

specification/examples/ShipmentResponse.json
```diff
@@ -6,6 +6,7 @@
     {
       "id": "7b808eee-bf1c-40cd-98f2-3c335a06417e",
       "attributes": {
+        "type": "object",
```
NickVries: This should not be here.

specification/schemas/PatchCarrierContract.json
```diff
@@ -0,0 +1,64 @@
...
+      "properties": {
+        "attributes": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            },
+            "active": {
+              "type": "boolean",
```
NickVries: You could add the default value here as well.

PR #201, opened 2018-05-28
:bug: removed additionalProperties

nessy:


specification/schemas/CarrierContract.json
```diff
@@ -12,7 +12,6 @@
       "properties": {
         "attributes": {
           "type": "object",
-          "additionalProperties": false,
```
Shodske: Why is this removed?

NickVries: With post carrier contract, we want to specify that you can send credentials. So that's done with an `"allOf"` in the post path. But `"allOf"` is not possible when `"additionalProperties": false` is present.

specification/paths/CarrierContracts.json
```diff
@@ -181,6 +192,17 @@
                             "currency",
                             "name"
                           ]
+                        },
+                        "relationships": {
+                          "properties": {
+                            "owner": {
+                              "example": {
+                                "links": {
```
Shodske: `links` shouldn't just be an example, it should be properly defined in a schema.

NickVries: The owner relationship is in this case a `"oneOf"` relationship. Which means that our UI doesn't automatically produce an example. If we put it in the example of the schema, that means that post also has links in the example. That's why it was put here as an extra for responses.

PR #206, opened 2018-06-08
Feature/swagger ui

M4tini:
Updated swagger-ui and the gray model-box css.
Fixed the `stringnull` rendering issue.

### DIFF
--- a/src/myparcelcom.js
+++ b/src/myparcelcom.js
@@ -55,7 +55,7 @@
   document.addEventListener('click', function () {
     // change 'stringnull' to 'string null'
     window.requestAnimationFrame(function () {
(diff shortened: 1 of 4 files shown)


src/myparcelcom.js
```diff
@@ -12,47 +12,56 @@
...
+          accordeons[i].addEventListener('click', cancelEvent)
         }
       }
     }, 137)
   })
+
+  document.addEventListener('click', function () {
+    // change 'stringnull' to 'string null'
+    window.requestAnimationFrame(function () {
+      let properties = document.getElementsByClassName('prop-type')
```
denniskoster: I think this can actually be a `const`. Values inside may change but the assignment remains the same.

PR #211, opened 2018-06-19
Feature/user status

M4tini:


specification/schemas/User.json
```diff
@@ -32,6 +32,15 @@
               "type": "string",
               "example": "+31 234 567 890"
             },
+            "status": {
+              "type": "string",
```
NickVries: You probably want to make this readOnly.

PR #239, opened 2018-08-28
✨ Invoices specification

Shodske:


specification/paths/Invoices-invoice_id.json
```diff
@@ -0,0 +1,171 @@
...
+    ],
+    "summary": "Update a invoice by id.",
+    "description": "This endpoint updates the invoice with the supplied id.",
+    "parameters": [
+      {
+        "$ref": "#/components/parameters/path-invoice_id"
+      }
+    ],
+    "requestBody": {
+      "description": "The organization object with attributes to be updated.",
```
M4tini: invoice

specification/paths/Invoices-invoice_id.json
```diff
@@ -0,0 +1,171 @@
...
+    "tags": [
+      "Invoices"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+        ]
+      }
+    ],
+    "summary": "Update a invoice by id.",
```
M4tini: an

specification/schemas/Invoice.json
```diff
@@ -0,0 +1,93 @@
...
+    {
+      "properties": {
+        "attributes": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "description": {
+              "oneOf": [
+                {
+                  "type": "string",
```
NickVries: You can use `"type":  ["string", "null"]`, instead of the oneOf when the type is a simple string or number.

specification/paths/Invoices-invoice_id.json
```diff
@@ -0,0 +1,171 @@
...
+      {
+        "OAuth2": [
+        ]
+      }
+    ],
+    "summary": "Get one invoice.",
+    "description": "",
+    "parameters": [
+      {
+        "$ref": "#/components/parameters/path-invoice_id"
```
NickVries: This parameter does't exist yet.

specification/schemas/Invoice.json
```diff
@@ -0,0 +1,93 @@
...
+                {
+                  "type": "null"
+                }
+              ]
+            },
+            "amount": {
+              "type": "number",
+              "example": 7985
+            },
+            "currency": {
```
M4tini: I am making the invoice transformer and am wondering why the `amount` and `currency` are separate properties. All other models have a `price` object including these two properties. Let's quickly discuss this after the standup.

PR #240, opened 2018-08-30
Add payment endpoints

nessy:
https://myparcelcombv.atlassian.net/browse/MP-1388


specification/paths/Payments-payment_id.json
```diff
@@ -0,0 +1,98 @@
...
+  "get": {
+    "tags": [
+      "Payments"
+    ],
+    "security": [
+      {
+        "OAuth2": []
+      }
+    ],
+    "summary": "Get an payment by id.",
```
Shodske: `"Get a payment by id."`

specification/schemas/MinimalPayment.json
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "object",
+  "required": [
+    "type"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
+      "pattern": "^Payments$",
+      "example": "Payments"
```
Shodske: This shouldn't be capitalized.

specification/schemas/InvoiceRelationship.json
```diff
@@ -0,0 +1,27 @@
...
+  "type": "object",
+  "required": [
+    "data"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "data": {
+      "type": "object",
+      "required": [
+        "type"
```
Shodske: `id` should also be required in a relationship, and the `related` link should be added as well.

specification/schemas/PatchPayment.json
```diff
@@ -0,0 +1,46 @@
...
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            }
+          }
+        },
+        "relationships": {
+          "type": "object",
+          "required": [
+            "organization",
+            "status"
```
Shodske: These shouldn't be required in a patch.

specification/schemas/PatchUser.json
```diff
@@ -14,11 +14,17 @@
           "additionalProperties": false,
           "properties": {
             "first_name": {
-              "type": "string",
+              "type": [
+                "string",
+                "null"
```
Shodske: I'd rather not have changes like this in a pull request that's unrelated to patching users. Did we discuss that first and last name should be nullable?

specification/schemas/Payment.json
```diff
@@ -0,0 +1,85 @@
...
+        "attributes": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "amount": {
+              "$ref": "#/components/schemas/PriceAmount"
+            },
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            },
```
Shodske: Should  we move `amount` and `currency` to a `price` property?

specification/schemas/Payment.json
```diff
@@ -0,0 +1,85 @@
...
+            "updated_at": {
+              "$ref": "#/components/schemas/Timestamp"
+            }
+          }
+        },
+        "relationships": {
+          "type": "object",
+          "required": [
+            "organization",
+            "status"
```
Shodske: These should not be required in the base model, as we want them not to be required in a patch.

specification/schemas/PaymentIdentityRelationship.json
```diff
@@ -0,0 +1,27 @@
...
+  "type": "object",
+  "required": [
+    "data"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "data": {
+      "type": "object",
+      "required": [
+        "type"
```
Shodske: `id` should be required and you should add a `related` link.

specification/schemas/StatusRelationship.json
```diff
@@ -1,5 +1,6 @@
 {
   "type": "object",
+  "readOnly": true,
```
Shodske: I don't think this should be read only. The payment application probably needs to be able to set the status by itself.

specification/schemas/InvoiceRelationship.json
nessy: Invoice does not have a endpoint yet for the related link

specification/schemas/PaymentIdentityRelationship.json
nessy: Payment Identity does not have a endpoint yet for the related link

Shodske: There should be an endpoint for it in the current spec.

specification/schemas/InvoiceRelationship.json
Shodske: That was added in v0.21.2

specification/schemas/PatchUser.json
nessy: This is already the case Nick was looking with me when we noticed it being wrong in the spec.

specification/paths/Payments-payment_id.json
```diff
@@ -0,0 +1,117 @@
...
+              "additionalProperties": false,
+              "properties": {
+                "data": {
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/Payment"
+                    },
+                    {
+                      "required": [
+                        "attributes"
```
Shodske: `id` and `relationships` should be required in the get endpoints.

specification/paths/Payments-payment_id.json
```diff
@@ -0,0 +1,117 @@
...
+                      ]
+                    },
+                    {
+                      "properties": {
+                        "relationships": {
+                          "required": [
+                            "organization",
+                            "status"
+                          ]
+                        }
```
Shodske: You should also add `attributes`, with a required array.

specification/paths/Payments.json
```diff
@@ -0,0 +1,162 @@
...
+                  "type": "array",
+                  "items": {
+                    "allOf": [
+                      {
+                        "$ref": "#/components/schemas/Payment"
+                      },
+                      {
+                        "required": [
+                          "id",
+                          "attributes"
```
Shodske: `relationships` should be required

specification/paths/Payments.json
```diff
@@ -0,0 +1,162 @@
...
+                        ]
+                      },
+                      {
+                        "properties": {
+                          "relationships": {
+                            "required": [
+                              "organization",
+                              "status"
+                            ]
+                          }
```
Shodske: `attributes` should be added with a required array

specification/paths/Payments.json
```diff
@@ -0,0 +1,162 @@
...
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "allOf": [
+                  {
+                    "$ref": "#/components/schemas/Payment"
+                  },
+                  {
+                    "required": [
+                      "attributes"
```
Shodske: `relationships` should be required

specification/paths/Payments.json
```diff
@@ -0,0 +1,162 @@
...
+                    ]
+                  },
+                  {
+                    "properties": {
+                      "relationships": {
+                        "required": [
+                          "organization",
+                          "status"
```
Shodske: I think (but I'm not certain) status can be removed as this can always be set to a default status when posting.

specification/paths/Payments.json
```diff
@@ -0,0 +1,162 @@
...
+                        ]
+                      }
```
Shodske: `attributes` should be added with a required array.

PR #246, opened 2018-09-09
:recycle: patch payment schema

M4tini:
- Added missing scope to payment identities.
- Limited payment patch attributes and relationships.

### DIFF
--- a/specification/schemas/PatchPayment.json
+++ b/specification/schemas/PatchPayment.json
@@ -13,11 +13,13 @@
           "type": "object",
           "additionalProperties": false,
           "properties": {
-            "amount": {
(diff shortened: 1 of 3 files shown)


specification/schemas/PatchPayment.json
```diff
@@ -13,27 +13,23 @@
...
-            "organization": {
-              "$ref": "#/components/schemas/OrganizationRelationship"
-            },
             "status": {
               "$ref": "#/components/schemas/StatusRelationship"
             },
-            "invoice": {
-              "$ref": "#/components/schemas/InvoiceRelationship"
-            },
             "payment_identity": {
```
NickVries: I think payment_identity should probably also not be patched right?

PR #247, opened 2018-09-10
Change carrier fields to not be required. Require created_at

denniskoster:
**Changes**
- `carrier_` fields should not be required in shipment status
- `created_at` should be required in shipment status

### DIFF
--- a/specification/schemas/ShipmentStatus.json
+++ b/specification/schemas/ShipmentStatus.json
@@ -12,9 +12,7 @@
         "attributes": {
           "type": "object",
           "required": [
-            "carrier_status_code",
-            "carrier_status_description",
-            "carrier_timestamp"
+            "created_at"
           ],
           "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


specification/schemas/ShipmentStatus.json
```diff
@@ -32,6 +30,11 @@
               "type": "number",
               "example": 1504801719,
               "description": "Unix timestamp when the carrier registered this status."
+            },
+            "created_at": {
+              "type": "number",
+              "example": 1504801719,
+              "description": "Unix timestamp when the status was created."
```
M4tini: For created_at and updated_at we use this everywhere:
```
"created_at": {
  "$ref": "#/components/schemas/Timestamp"
}
```

denniskoster: We should probably use that for `carrier_timestamp` as well then. Agreed?

PR #257, opened 2018-10-09
✨ Removed active attribute from carrier contracts and updated status attribute

Shodske:


Shodske:
Please wait with merging until I've finalized the API as well, so people won't have failing tests.

PR #273, opened 2018-11-08
Feature/mp 1588 service contract rework

SAMDevelopment:


denniskoster:
@Shodske Looks good to me, but I don't have the clearest vision of the complete rework. Could you take a look as well?

PR #290, opened 2019-01-03
Add client resource and endpoints to spec

denniskoster:
https://myparcelcombv.atlassian.net/browse/MP-1889

### DIFF
--- a/specification/paths/Clients-client_id.json
+++ b/specification/paths/Clients-client_id.json
@@ -1,0 +1,118 @@
+{
+  "get": {
+    "tags": [
+      "Clients"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "organizations.manage"
+        ]
+      }
+    ],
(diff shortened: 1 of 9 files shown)


specification/paths/Clients-client_id.json
```diff
@@ -0,0 +1,118 @@
+{
+  "get": {
+    "tags": [
+      "Clients"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "organizations.manage"
```
Shodske: I don't think we need a scope to create a client at the moment, do we?

denniskoster: Well, at the moment you can't really create a client. The task describes we should add crud permissions for clients and add them to the organizations.manage scope, so that's what I did.

Shodske: I see, fair enough then.

PR #310, opened 2019-02-26
:recycle: added created_at to client resource

M4tini:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-1557

### DIFF
--- a/specification/schemas/Client.json
+++ b/specification/schemas/Client.json
@@ -22,6 +22,9 @@
             "label": {
               "type": "string",
               "example": "Magento"
+            },
+            "created_at": {
+              "$ref": "#/components/schemas/Timestamp"
             }
           }
         }


specification/schemas/Client.json
```diff
@@ -22,6 +22,14 @@
             "label": {
               "type": "string",
               "example": "Magento"
+            },
+            "grant_type": {
```
Shodske: I believe we didn't add this yet as we will not allow customers to create other clients than `client_credentials`. Any reason you have added it now?

M4tini: I added it as a read-only field, because I want to hide the delete button for our `implicit` grant tasks. As discussed, we will change this to the API only returning `client_credentials` clients.

PR #313, opened 2019-03-05
[MP-1555] Add client origin meta to shipment

fernandofreamunde:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-1555

Not sure if I need to add something else...
Also the ticked mentions the relationship name should be `origin` but has a question mark. can we discuss that here? 

I'll implement whatever we decide later :)

### DIFF
--- a/specification/schemas/InternationalShipment.json
+++ b/specification/schemas/InternationalShipment.json
@@ -210,6 +210,17 @@
             }
           }
         },
+        "meta": {
+          "readOnly": true,
+          "type": "object",
+          "properties": {
+            "origin": {
+              "type": "string",
+              "example": "MyParcel.com backoffice",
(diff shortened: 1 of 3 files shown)


specification/schemas/InternationalShipment.json
```diff
@@ -210,6 +210,17 @@
...
           }
         },
+        "meta": {
+          "readOnly": true,
+          "type": "object",
+          "properties": {
+            "origin": {
+              "type": "string",
+              "example": "MyParcel.com backoffice",
+              "description": "Reference of application client which the shipment was created."
```
denniskoster:
```suggestion
              "description": "Label of the API client used to create the shipment."
```

PR #315, opened 2019-03-13
MP-2175 Added /suggest-address RPC endpoint

NickVries:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2175

### DIFF
--- a/specification/paths/SuggestAddress.json
+++ b/specification/paths/SuggestAddress.json
@@ -40,6 +40,10 @@
                   "street_number": {
                     "type": "string",
                     "example": "679"
+                  },
+                  "street_number_suffix": {
+                    "type": "string",
+                    "example": "A1"
                   }
                 }
               }
@@ -76,8 +80,8 @@
                         "example": "2131 BC"
                       },
                       "street_number": {
-                        "type": "number",
-                        "example": 679
+                        "type": "string",
+                        "example": "679"
                       },
                       "city": {
                         "type": "string",
--- a/specification/paths/SuggestAddress.json
+++ b/specification/paths/SuggestAddress.json
@@ -6,7 +6,7 @@
     "security": [
       {
         "OAuth2": [
-          "addresses:suggest"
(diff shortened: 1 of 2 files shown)


specification/paths/SuggestAddress.json
```diff
@@ -0,0 +1,104 @@
+{
+  "post": {
+    "tags": [
+      "RPC"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "addresses:suggest"
```
Shodske: Should be `addresses.suggest`

specification/paths/SuggestAddress.json
```diff
@@ -0,0 +1,104 @@
...
+                  "country_code": {
+                    "type": "string",
+                    "example": "NL"
+                  },
+                  "postal_code": {
+                    "type": "string",
+                    "example": "2131 BC"
+                  },
+                  "street_number": {
+                    "type": "string",
```
Shodske: Should be a number and I think we should add `street_number_suffix` as a string.

NickVries: After discussing this with the team, we decided to leave the `street_number` property as a string (some street numbers include letters: I live on street number 3A with suffix 1).
In order to be consistent we should probably change types of the street number properties in other resources to string as well, but that will break things and seems like a separate issue.

I did add the `street_number_suffix` property.

Shodske: I think that keeping it a string will cause more problems than that are solved by it. In your case I would assume that you live on number 3 with suffix A-1 (or A1, A/1, etc). If we allow any string to be entered we would always have to check that no suffixes are present in the number.

Another thing is that the PostNL API also requires the street number to be a number. Which means that if we were to take your case with `"street_numer": "3A"` and `"street_number_suffix": "1"`, we would still have to strip the `A` from the number and add it to the `1` from the suffix. So separating these two fields would have no benefit for us if we don't put constraints on them.

PR #321, opened 2019-04-16
Feature/mp 2189 use integer instead of number

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-2189
---
Instead of only the weight, I changed all `number` which may not be a float to `integer`.

:tada: Running the API tests with this new spec is successful.
:ok_hand: Creating a shipment with weight `0.350` now results in a `400` spec error with:
```
{
    "property": "data.attributes.physical_properties.weight",
    "pointer": "/data/attributes/physical_properties/weight",
    "message": "Double value found, but an integer is required",
    "constraint": "type",
    "context": 1
}
```

### DIFF
--- a/specification/schemas/PatchServiceRate.json
+++ b/specification/schemas/PatchServiceRate.json
@@ -14,33 +14,33 @@
           "additionalProperties": false,
           "properties": {
             "weight_min": {
-              "type": "number",
+              "type": "integer",
               "example": 0,
               "description": "Weight in grams."
             },
             "weight_max": {
-              "type": "number",
+              "type": "integer",
               "example": 2000,
               "description": "Weight in grams."
             },
             "length_max": {
-              "type": "number",
+              "type": "integer",
               "example": 300,
               "description": "Length in mm."
             },
             "width_max": {
-              "type": "number",
+              "type": "integer",
               "example": 200,
               "description": "Width in mm."
             },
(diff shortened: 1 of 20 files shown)


specification/schemas/ServiceRate.json
```diff
@@ -18,33 +18,33 @@
...
               "description": "Width in mm."
             },
             "height_max": {
-              "type": "number",
+              "type": "integer",
               "example": 200,
               "description": "Height in mm."
             },
             "volume_max": {
               "type": "number",
```
nessy: Should this not be float if it can contain non whole numbers?

specification/schemas/PatchServiceRate.json
```diff
@@ -14,33 +14,33 @@
...
               "description": "Width in mm."
             },
             "height_max": {
-              "type": "number",
+              "type": "integer",
               "example": 200,
               "description": "Height in mm."
             },
             "volume_max": {
               "type": "number",
```
nessy: Should this not be float if it can contain non whole numbers?

specification/schemas/ServiceRate.json
M4tini: `float` is not a JSON Schema numeric value:
https://json-schema.org/understanding-json-schema/reference/numeric.html

PR #322, opened 2019-04-19
:sparkles: Added fields city, street_1 and renamed response.street_na

yoan-myparcel:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2315

### DIFF
--- a/specification/paths/SuggestAddress.json
+++ b/specification/paths/SuggestAddress.json
@@ -25,9 +25,6 @@
             "properties": {
               "data": {
                 "type": "object",
-                "required": [
-                  "country_code"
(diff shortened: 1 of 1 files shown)


specification/paths/SuggestAddress.json
```diff
@@ -25,9 +25,6 @@
             "properties": {
               "data": {
                 "type": "object",
-                "required": [
-                  "country_code"
-                ],
```
yoan-myparcel: FYI I am not entirely sure if it is possible to do comprehensive required rules like with Laravel, therefore I removed the required constraint here.

PR #330, opened 2019-05-21
Add hook logs

nessy:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2307

### DIFF
--- a/specification/paths/HookLogs-hook_log_id.json
+++ b/specification/paths/HookLogs-hook_log_id.json
@@ -30,17 +30,7 @@
               "additionalProperties": false,
               "properties": {
                 "data": {
-                  "allOf": [
-                    {
-                      "$ref": "#/components/schemas/HookLog"
-                    },
-                    {
-                      "required": [
-                        "id",
-                        "links"
-                      ]
-                    }
-                  ]
+                  "$ref": "#/components/schemas/HookLog"
                 }
               }
             }
--- a/specification/paths/HookLogs-hook_log_id.json
+++ b/specification/paths/HookLogs-hook_log_id.json
@@ -1,0 +1,52 @@
+{
+  "get": {
+    "tags": [
+      "HookLogs"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "experimental"
+        ]
+      }
+    ],
+    "summary": "Get a hook log by id.",
+    "description": "This endpoint retrieves the hook log with the supplied id.",
+    "parameters": [
+      {
+        "$ref": "#/components/parameters/path-hook-log_id"
+      }
+    ],
+    "responses": {
+      "200": {
+        "description": "Retrieved the hook log.",
+        "content": {
+          "application/vnd.api+json": {
+            "schema": {
+              "type": "object",
+              "required": [
+                "data"
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "data": {
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/HookLog"
+                    },
+                    {
+                      "required": [
+                        "id",
+                        "links"
+                      ]
+                    }
+                  ]
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
+{
+  "type": "object",
+  "required": [
+    "attributes",
```
NickVries: You can add "id" and "links" in here since they will always be there. We will never add a post hooklogs endpoint so there will never be a case that we will not have an id or links for this resource.

See above comment.

specification/paths/HookLogs-hook_log_id.json
```diff
@@ -0,0 +1,52 @@
...
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "data": {
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/HookLog"
+                    },
+                    {
+                      "required": [
```
NickVries: You can add "id" and "links" into the schema since they will always be there. We will never add a post hooklogs endpoint so there will never be a case that we will not have an id or links for this resource.

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
...
+    "relationships"
+  ],
+  "properties": {
```
NickVries: Where are the `type` and `id` properties?? 

image

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
...
+    "attributes": {
+      "type": "object",
+      "required": [
+        "resource_diff",
+        "errors"
```
NickVries: Are you sure we will always have a resource_diff and errors?

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
...
+              }
+            }
+          },
+          "description": "The difference between the resource before and after the hook has run."
+        },
+        "errors": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "example": "Applying the hook failed because the selecting service was not available for the shipment contract."
```
NickVries:
```suggestion
            "example": "Applying the hook failed because the desired service was not available for the shipment contract."
```

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
...
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "resource_diff": {
+          "type": "object",
+          "example": {
+            "relationships" : {
+              "service": {
+                "data": {
+                  "id" : "7b808eee-bf1c-40cd-98f2-3c335a06417e"
```
NickVries: Maybe it's a good idea to also put the `type` property with value `services` in here, because that's what the hooks usually also set. There's no use in setting just an `id` or just a `type`.

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,121 @@
...
+              "type": "shipments",
+              "id": "9cdf86e8-333f-4ed9-bb31-4935c780c947"
+            }
+          }
+        },
+        "hook": {
+          "$ref": "#/components/schemas/HookLogRelationship",
+          "example": {
+            "data": {
+              "type": "organizations",
```
NickVries: Doesn't the HookLogRelationship schema already specify an example? And also `"type": "organizations"`??

specification/schemas/HookLogRelationship.json
```diff
@@ -0,0 +1,42 @@
...
+  "type": "object",
+  "required": [
+    "data"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "data": {
+      "type": "object",
+      "required": [
+        "type"
```
NickVries: Shouldn't `id` also be required?

specification/schemas/HookLog.json
```diff
@@ -0,0 +1,125 @@
...
+          },
+          "example": {
+            "data": {
+              "type": "shipments",
+              "id": "9cdf86e8-333f-4ed9-bb31-4935c780c947"
+            }
+          }
+        },
+        "hook": {
+          "$ref": "#/components/schemas/HookLogRelationship"
```
NickVries: This should be the a HookRelationship schema, not the HookLogRelationship.

PR #349, opened 2019-06-19
Remove one of international shipment

nessy:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2484


specification/paths/Shipments.json
```diff
@@ -78,71 +78,33 @@
...
-                                ]
-                              },
-                              "relationships": {
-                                "required": [
-                                  "shop",
-                                  "shipment_status"
-                                ]
-                              }
-                            }
+                              "shop",
```
NickVries: `shop` is also already in the required relationships in the Shipment schema.

specification/paths/Shipments.json
```diff
@@ -217,34 +179,16 @@
             "additionalProperties": false,
             "properties": {
               "data": {
-                "oneOf": [
+                "allOf": [
```
NickVries: You can remove this allOf completely, since the part with `id` readOnly true is already in the Shipment schema.

NickVries:
I've removed the example request bodies and responses for shipments. They were initially added because the "oneOf" constraints caused the spec to not be able to generate an example because it didn't know what to show. Since we've removed the oneOf constraints, we can also remove the examples. I've compared the examples with a comparison tool and it all seems fine. Either @M4tini or @yoan-myparcel should probably go over the PR and approve, since @nessy and I have both contributed.

PR #369, opened 2019-08-12
:recycle: add channel and total_value to patch shipment schema

M4tini:


specification/schemas/PatchShipment.json
```diff
@@ -49,6 +49,14 @@
               ],
               "example": "Order #8008135"
             },
+            "channel": {
+              "type": "string",
+              "example": "MyParcel.com backoffice",
+              "description": "Name of the application used to create the shipment."
+            },
+            "total_value": {
+              "$ref": "#/components/schemas/Price"
```
NickVries: Do we have a BasePrice schema to make sure not both value and currency are required when you only want to change the value?

PR #377, opened 2019-08-29
🐛 Using the same validation criteria as in RFC 1123 

yoan-myparcel:
**Changes**
- I looked into both how Aurelia and the api-specification validates e-mail and the API spec is more restrictive. I copied the regex that is used by Aurelia. 

**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2633


specification/schemas/Email.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "string",
-  "pattern": "^(([^<>()\\[\\]\\.,;:\\s@\"]+(\\.[^<>()\\[\\]\\.,;:\\s@\"]+)*)|(\".+\"))@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$",
+  "pattern": "/^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/",
```
M4tini: I think the surrounding `/` characters need to be removed.
Also, regex101 gives me a pattern error (for both regular expressions).
Furthermore, the regex in the backoffice is too lenient, as it validates emails without an extension: `dont@mail`

Let's use the regex from https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address in both the spec and the backoffice. You can copy the custom ValidationRule in `src/backoffice.js`.

yoan-myparcel:
```suggestion
  "pattern": "^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$",
```

PR #379, opened 2019-08-30
✨ Added shipment surcharges to specification 

yoan-myparcel:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2700
- https://myparcelcombv.atlassian.net/browse/MP-2669

### DIFF
--- a/specification/paths/ShipmentSurcharges.json
+++ b/specification/paths/ShipmentSurcharges.json
@@ -1,4 +1,105 @@
 {
+  "get": {
+    "tags": [
+      "ShipmentSurcharges"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "finance.manage"
+        ]
+      }
+    ],
+    "summary": "Get all surcharges across all shipments.",
+    "description": "This endpoint retrieves all surcharges for all shipments.",
+    "parameters": [
+      {
+        "name": "include",
+        "in": "query",
+        "description": "Comma separated string of the relationship names you want to include the data of.<br>The relationships that can be included are:<ul><li>`shipment`</li></ul>",
+        "schema": {
+          "type": "string"
+        }
+      }
+    ],
+    "responses": {
+      "200": {
+        "description": "Retrieved all of the shipment surcharges.",
+        "content": {
+          "application/vnd.api+json": {
+            "schema": {
+              "type": "object",
+              "required": [
+                "data",
+                "meta",
+                "links"
+              ],
+              "properties": {
+                "data": {
+                  "type": "array",
+                  "items": {
+                    "allOf": [
+                      {
+                        "$ref": "#/components/schemas/ShipmentSurcharge"
+                      },
+                      {
+                        "required": [
+                          "id",
+                          "links"
+                        ],
+                        "properties": {
+                          "attributes": {
+                            "required": [
+                              "created_at",
+                              "updated_at"
+                            ]
+                          }
+                        }
+                      }
+                    ]
+                  }
+                },
+                "meta": {
+                  "$ref": "#/components/schemas/PaginationMeta"
+                },
+                "links": {
+                  "type": "object",
+                  "required": [
+                    "self",
+                    "first",
+                    "last"
+                  ],
+                  "additionalProperties": false,
+                  "properties": {
+                    "self": {
+                      "type": "string",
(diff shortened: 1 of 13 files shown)


specification/schemas/PatchShipmentSurcharge.json
```diff
@@ -0,0 +1,56 @@
...
+              "type": "integer",
+              "example": 523,
+              "description": "Surcharge amount in cents/pence."
+            },
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            }
+          }
+        },
+        "links": {
```
NickVries: You can leave out the `links`, since you won't be able to patch them.

specification/schemas/ShipmentSurcharge.json
```diff
@@ -0,0 +1,82 @@
...
+          ],
+          "additionalProperties": false,
+          "properties": {
+            "shipment": {
+              "allOf": [
+                {
+                  "$ref": "#/components/schemas/ShipmentRelationship"
+                },
+                {
+                  "readOnly": false
```
NickVries: Is the ShipmentRelationship "readOnly"?

yoan-myparcel: If readOnly is set to `true` the example request will be empty in the specs page.

NickVries: No I understand that, I was just wondering if the ShipmentRelationship schema includes `"readOnly": true`. Since you're trying to "change" it to false here.

yoan-myparcel: It is readOnly: true

specification/schemas/PatchShipmentSurcharge.json
```diff
@@ -0,0 +1,41 @@
...
+            "description": {
+              "type": "string",
+              "example": "A carrier-specific (carrier_code) re-labeling surcharge"
+            },
+            "amount": {
+              "type": "integer",
+              "example": 523,
+              "description": "Surcharge amount in cents/pence."
+            },
+            "currency": {
```
M4tini: The `amount` and `currency` should be replaced with:
```
"price": {
  "$ref": "#/components/schemas/BasePrice"
},
```

specification/schemas/ShipmentSurcharge.json
```diff
@@ -0,0 +1,82 @@
...
+            "description": {
+              "type": "string",
+              "example": "A carrier-specific (carrier_code) re-labeling surcharge"
+            },
+            "amount": {
+              "type": "integer",
+              "example": 523,
+              "description": "Surcharge amount in cents/pence."
+            },
+            "currency": {
```
M4tini: The `amount` and `currency` should be replaced with:
```
"price": {
  "$ref": "#/components/schemas/Price"
},
```

specification/schemas/PatchShipmentSurcharge.json
yoan-myparcel: Gourav here. I thought about doing that at first. However, `price` is not an attribute name that would match the context of Surcharges, isn't it? It is not really a price we are talking about but exactly an amount that has currency.

specification/paths/ShipmentSurcharges.json
```diff
@@ -0,0 +1,147 @@
...
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "finance.manage"
+        ]
+      }
+    ],
+    "summary": "Get all surcharges across all shipments.",
+    "description": "This endpoint retrieves surcharges for the supplied shipment.",
```
NickVries:
```suggestion
    "description": "This endpoint retrieves all surcharges for all shipments.",
```

specification/schemas/PatchShipmentSurcharge.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "required": [
+        "attributes"
+      ],
+      "properties": {
+        "attributes": {
+          "type": "object",
+          "required": [
+            "amount",
+            "currency"
```
NickVries: Why are both amount and currency required when making a change here? Shouldn't you be able to just send the amount if the currency stays the same?

specification/schemas/PatchShipmentSurcharge.json
M4tini: Agreed. We can think of another name to use for this combined property. In other occasions we have used `item_value`, `total_value` and `purchase_price`. What about `surcharge`?

PR #391, opened 2019-09-19
Add information to services

nessy:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2755

### DIFF
--- a/specification/schemas/Service.json
+++ b/specification/schemas/Service.json
@@ -38,10 +38,10 @@
               "example": [
                 {
                   "type": "notice",
-                  "text": "Volumetric weight may be applied."
+                  "body": "Volumetric weight may be applied."
                 }
               ],
-              "description": "Additional information about the service"
+              "description": "Additional information about the service."
             },
             "package_type": {
               "type": "string",
--- a/specification/schemas/Service.json
+++ b/specification/schemas/Service.json
@@ -33,6 +33,16 @@
               "example": "spring-pap",
               "description": "Slug of the service"
             },
+            "information": {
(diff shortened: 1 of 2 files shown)


specification/schemas/Service.json
```diff
@@ -33,6 +33,16 @@
...
             },
+            "information": {
+              "type": "array",
+              "example": [
+                {
+                  "type": "notice",
+                  "text": "Volumetric weight may be applied."
+                }
+              ],
+              "description": "Additional information about the service"
```
NickVries:
```suggestion
              "description": "Additional information about the service."
```

specification/schemas/Service.json
```diff
@@ -33,6 +33,16 @@
               "example": "spring-pap",
               "description": "Slug of the service"
             },
+            "information": {
+              "type": "array",
+              "example": [
+                {
+                  "type": "notice",
+                  "text": "Volumetric weight may be applied."
```
NickVries: I would maybe call this `body` instead of `text`.

specification/schemas/Service.json
```diff
@@ -33,6 +33,16 @@
               "example": "spring-pap",
               "description": "Slug of the service"
             },
+            "information": {
+              "type": "array",
+              "example": [
+                {
+                  "type": "notice",
```
NickVries: Do you have an idea on how many different `types` there will be? Maybe we can make this an enum?

nessy: We don't know the rest of the types, so we will just add them later

PR #407, opened 2019-10-28
Add shipment consolidation relation to shipments

nessy:
**Related issues**
- https://myparcelcombv.atlassian.net/browse/MP-2853

### DIFF
--- a/specification/schemas/Shipment.json
+++ b/specification/schemas/Shipment.json
@@ -118,6 +118,12 @@
                 "$ref": "#/components/schemas/ShipmentItem"
               }
             },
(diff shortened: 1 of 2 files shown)


specification/schemas/Shipment.json
```diff
@@ -118,6 +118,12 @@
                 "$ref": "#/components/schemas/ShipmentItem"
               }
             },
+            "is_consolidation": {
+              "readOnly": true,
```
M4tini: We need to be able to set this via the backoffice, so it should not be read-only.

PR #424, opened 2019-12-10
MP-2231 ♻️ Add uuid regex

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-2231
I already tested this fix by copying the spec to my local API.

### DIFF
--- a/specification/paths/SynchronizeIntegration.json
+++ b/specification/paths/SynchronizeIntegration.json
@@ -33,11 +33,13 @@
                   "shop_id": {
                     "type": "string",
                     "format": "uuid",
+                    "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
                     "example": "35eddf50-1d84-47a3-8479-6bfda729cd99"
                   },
                   "integration_id": {
                     "type": "string",
                     "format": "uuid",
+                    "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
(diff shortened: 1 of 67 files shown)


specification/schemas/Email.json
```diff
@@ -1,5 +1,6 @@
 {
   "type": "string",
+  "format": "email",
```
yoan-myparcel: If we have the pattern what value does the format bring?

specification/paths/SynchronizeIntegration.json
```diff
@@ -33,11 +33,13 @@
                   "shop_id": {
                     "type": "string",
                     "format": "uuid",
+                    "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
```
yoan-myparcel: we cannot use the Uuid schema here?

specification/schemas/Email.json
M4tini: Nothing, just for clarification. Maybe they eventually update the json-api package to mention the `format` when the `pattern` does not match.

specification/paths/SynchronizeIntegration.json
M4tini: The problem with the Uuid schema is that it does not include an example. And we do not want the same example uuid for all our resources. That would look weird in relationship examples.

Now I think about it, we could use the Uuid schema like this:
```
"allOf": [
  {
  },
  {
    "example": "35eddf50-1d84-47a3-8479-6bfda729cd99"
  }
]
```

PR #435, opened 2020-02-27
✨ Void shipment endpoint

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-2914
- `/void-shipment` RPC endpoint accepting shipment_id in post data.
- Returning 204 on success and 423 on error.

### DIFF
--- a/specification/info.json
+++ b/specification/info.json
@@ -11,5 +11,5 @@
     "name": "MyParcel.com general terms and conditions",
     "url": "https://www.myparcel.com/terms"
   },
-  "version": "1.15.0"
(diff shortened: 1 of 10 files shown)


specification/info.json
```diff
@@ -11,5 +11,5 @@
     "name": "MyParcel.com general terms and conditions",
     "url": "https://www.myparcel.com/terms"
   },
-  "version": "1.15.0"
+  "version": "1.17.0"
```
yoan-myparcel: Perhaps it is time to automate the versioning on this one.

PR #440, opened 2020-03-26
MP-3211 Added customer reference to shipment attributes.

NickVries:
**Related Issues**
- [Subtask](https://myparcelcombv.atlassian.net/browse/MP-3211)
- [Story](https://myparcelcombv.atlassian.net/browse/MP-3155)

### DIFF
--- a/specification/paths/Shipments.json
+++ b/specification/paths/Shipments.json
@@ -24,7 +24,15 @@
       {
         "name": "filter[search]",
         "in": "query",
-        "description": "String of characters you want to look for in a shipment's attributes. Available fields to look in are:<ul><li>`date_created` (YYYY-MM-DD). <strong>Deprecated</strong> use `filter[date_from]` and `filter[date_to]` instead.</li><li>`description`</li><li>`recipient_address`</li><li>`tracking_code`</li></ul>",
+        "description": "String of characters you want to look for in a shipment's attributes. Available fields to look in are:<ul><li>`date_created` (YYYY-MM-DD). <strong>Deprecated</strong> use `filter[date_from]` and `filter[date_to]` instead.</li><li>`description`</li><li>`customer_reference`</li><li>`recipient_address`</li><li>`tracking_code`</li></ul>",
+        "schema": {
+          "type": "string"
+        }
+      },
+      {
+        "name": "filter[customer_reference]",
+        "in": "query",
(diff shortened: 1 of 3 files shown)


specification/paths/Shipments.json
```diff
@@ -24,7 +24,15 @@
...
-        "description": "String of characters you want to look for in a shipment's attributes. Available fields to look in are:<ul><li>`date_created` (YYYY-MM-DD). <strong>Deprecated</strong> use `filter[date_from]` and `filter[date_to]` instead.</li><li>`description`</li><li>`recipient_address`</li><li>`tracking_code`</li></ul>",
+        "description": "String of characters you want to look for in a shipment's attributes. Available fields to look in are:<ul><li>`date_created` (YYYY-MM-DD). <strong>Deprecated</strong> use `filter[date_from]` and `filter[date_to]` instead.</li><li>`description`</li><li>`customer_reference`</li><li>`recipient_address`</li><li>`tracking_code`</li></ul>",
+        "schema": {
+          "type": "string"
+        }
+      },
+      {
+        "name": "filter[customer_reference]",
+        "in": "query",
+        "description": "Search for a shipment by its customer reference attribute. If a shipments `customer_reference` attribute contains the search string, it is returned in the response.",
```
M4tini: I think we should not search `LIKE` because that's expensive.
My idea was to treat this field as an `id` so it should be an exact match.

specification/schemas/PatchShipment.json
```diff
@@ -51,7 +51,16 @@
...
-              "example": "Order #8008135"
+              "example": "Fidget spinners",
+              "description": "Description of the shipment (contents). This field is communicated to the carrier and may be put on the shipment's label."
+            },
+            "customer_reference": {
+              "type": [
+                "string",
+                "null"
+              ],
+              "example": "Order #8008135",
```
M4tini: Let's remove the `Order ` part, so we imply to just put the order number or ID here.

specification/schemas/Shipment.json
```diff
@@ -69,7 +69,16 @@
...
-              "example": "Order #8008135"
+              "example": "Fidget spinners",
+              "description": "Description of the shipment (contents). This field is communicated to the carrier and may be put on the shipment's label."
+            },
+            "customer_reference": {
+              "type": [
+                "string",
+                "null"
+              ],
+              "example": "Order #8008135",
```
M4tini: Let's remove the `Order ` part, so we imply to just put the order number or ID here.

PR #449, opened 2020-04-24
MP-3232 💥 broker_id instead of client_id in user meta

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3232

### DIFF
--- a/specification/paths/Users.json
+++ b/specification/paths/Users.json
@@ -133,14 +133,14 @@
               "meta": {
                 "type": "object",
                 "required": [
-                  "client_id"
+                  "broker_id"
                 ],
                 "properties": {
-                  "client_id": {
+                  "broker_id": {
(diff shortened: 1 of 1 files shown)


specification/paths/Users.json
```diff
@@ -133,14 +133,14 @@
               "meta": {
                 "type": "object",
                 "required": [
-                  "client_id"
+                  "broker_id"
                 ],
                 "properties": {
                   "client_id": {
```
yoan-myparcel, line 139:
```suggestion
                  "broker_id": {
```

yoan-myparcel, line 139:
```suggestion
                  "broker_id": {
```

M4tini: Oops :sweat_smile: This worked because additional properties are allowed.

PR #452, opened 2020-05-07
RPC /export-shipments

yoan-myparcel:
# Changes
Added new endpoint for downloading shipments in CSV details

### DIFF
--- a/specification/paths/ExportShipments.json
+++ b/specification/paths/ExportShipments.json
@@ -1,0 +1,118 @@
+{
+  "get": {
+    "tags": [
+      "RPC"
+    ],
+    "security": [
+      {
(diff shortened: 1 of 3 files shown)


specification/paths/ExportShipments.json
```diff
@@ -0,0 +1,118 @@
...
+        }
+      },
+      {
+        "name": "filter[organization]",
+        "in": "query",
+        "description": "Comma separated string of organization ids to filter by.",
+        "schema": {
+          "type": "string"
+        }
+      },
```
M4tini, line 63: I see you added `filter[organization]` to the export endpoint. That one is not yet available for the shipments endpoint, or does your API change make it available there as well?

PR #467, opened 2020-06-08
⚡️ Validation and navigation

M4tini:
- Filtered list of validation errors to report.
- Added missing scopes.
- Moved shared `path` parameters to path object.
- Fixed typo in `organization.manage` scope.
- Removed `deprecated` labels (now added by UI).
- Refactored redoc initialization to:
  - use callback instead of interval
  - scroll active menu item into view
- Marked experimental endpoints with ⚠
- Hide internal endpoints.
- Added tags to sort sections and display links to docs.

### DIFF
--- a/.validaterc
+++ b/.validaterc
@@ -3,13 +3,28 @@
     "operations": {
       "no_operation_id": "off"
     },
+    "parameters": {
+      "param_name_case_convention": "off",
+      "accept_type_parameter": "off"
+    },
+    "paths": {
+      "paths_case_convention": "off"
+    },
     "schemas": {
+      "invalid_type_format_pair": "off",
       "no_schema_description": "off",
       "no_property_description": "off",
-      "inconsistent_property_type": "off"
+      "description_mentions_json": "off",
+      "inconsistent_property_type": "off",
+      "enum_case_convention": "off"
     },
     "walker": {
       "no_empty_descriptions": "off"
+    }
+  },
+  "oas3": {
+    "operations": {
+      "no_request_body_name": "off"
     }
   }
 }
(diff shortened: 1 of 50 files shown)


src/myparcelcom.css
```diff
@@ -8,6 +8,12 @@ img[alt=logo] {
   margin: 20px 0 !important;
 }
 
-redoc {
-  display: block;
+/* avoid long regex to overlap the right area */
+.api-content > div > div:first-child > div:first-child {
+  overflow-x: scroll;
```
yoan-myparcel, line 13: I believe that this breaks the design somehow (check the x-scrolls on the image below):

image

.validaterc
```diff
@@ -3,13 +3,28 @@
...
     "walker": {
       "no_empty_descriptions": "off"
     }
+  },
+  "oas3": {
+    "operations": {
+      "no_request_body_name": "off"
+    }
   }
 }
```
yoan-myparcel, line 30: It makes sense to run the validator on CircleCI as a test job. However, when I run the validator locally it brings a lot of errors

src/myparcelcom.css
M4tini: Will change it to `auto` so Chrome does not show bars by default.

.validaterc
M4tini: It does, but at the moment we are using an OAS 3.1 feature to define fields as possibly nullable. OAS is currently at 3.0.3 so most errors will disappear when 3.1 is released. I tried turning off all possible rules in `.validaterc` but there are still errors because of this.

PR #476, opened 2020-07-14
MP-3280 🔥 removed users:manage scope from user id endpoints

M4tini:
Otherwise all users need to get `users.manage` on their user resource.


NickVries:
Introducing the experimental scope again on this endpoint kind of means that we still don't have a solid implementation for this. I would prefer to think of a good solution instead of introducing the experimental scope on one of our core endpoints.

PR #491, opened 2020-09-17
MP-3645 ✨ Post user scopes endpoint

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3536
I removed some duplicate user-scopes code in the process.

### DIFF
--- a/specification/schemas/ScopesRelationship.json
+++ b/specification/schemas/ScopesRelationship.json
@@ -13,7 +13,7 @@
           },
           {
             "required": [
-              "meta"
+              "id"
             ],
             "properties": {
               "meta": {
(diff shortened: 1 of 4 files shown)


specification/schemas/ScopesRelationship.json
```diff
@@ -13,7 +13,7 @@
           },
           {
             "required": [
-              "meta"
+              "id"
```
NickVries, line 16: Isn't `id` already required in the MinimalScope schema?

NickVries (changes requested):

specification/schemas/ScopesRelationship.json
M4tini: No, `minimal` schema's have only `type` because they can be used for PATCH requests where the ID should not be included.

NickVries: Gotcha :ok_hand:.

NickVries (approved):

PR #494, opened 2020-09-30
✨ Introduced shipment surcharges relationship and include

yoan-myparcel:
# Changes
- Introduced ShipmentSurchargeRelationship schema, updated includes


specification/schemas/Shipment.json
```diff
@@ -255,6 +255,30 @@
                 }
               ]
             },
+            "shipment_surcharges": {
+              "allOf": [
+                {
+                  "$ref": "#/components/schemas/ShipmentSurchargeRelationship"
```
NickVries, line 261: The `ShipmentSurchargeRelationship` schema describes a single resource with its own `data` attribute.
We need to also create a ShipmentSurcharge**s**Relationship that has a single `data` property, with multiple `MinimalShipmentSurcharge` schemas (just `type` and `id`).

NickVries (changes requested):

NickVries (approved):

PR #503, opened 2020-10-22
♻️ Removed the need for Patch schema's

M4tini:
## The new approach

- `ModelResource` = object with type and id, type is required (just rename the `MinimalModel` file).
- `Model` = object extending `ModelResource` with all attributes and relationships, nothing required, nullable if needed.
- `ModelResponse` = object extending `Model` with required attributes, relationships and links.

This approach removes the need for separate `Patch` models and response `allOf` with a lot of duplicate code.
It also fixes small stuff like the `id` not being marked as required for a GET response, etc.

This PR shows the changes I did for `Shops`, `SystemMessages` and `Users`.
The main reason behind this is to make `POST` endpoints accept `null` for optional attributes and relationships.


M4tini:
The API tests work fine now.

The "id in post example" bug is just a ReDoc UI issue ([and already existed before this PR](https://api-specification.myparcel.com/#tag/Shops/paths/~1shops/post)). I "fixed" it with some ugly JavaScript which we should delete as soon as they fix their codebase.

NickVries (approved):

PR #506, opened 2020-11-03
MP-3739 Added new analytics endpoint for carrier volumes

yoan-myparcel:
# Changes
Introduced new endpoint for volumes per carrier. All shipment filters are applicable.


M4tini (changes requested):
It seems like we duplicated the same set of shipment filters for 4 endpoints. Can we put the shared `filter[...]` stuff in a `ShipmentFilters.json` file and use `AllOf` to merge them with the 1 or 2 unique parameters like `interval`?

M4tini (approved):

PR #510, opened 2020-11-11
✨ Registered shipments endpoint

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3770

Leave the branch, I need to make a merge PR to develop.

### DIFF
--- a/specification/paths.json
+++ b/specification/paths.json
@@ -113,8 +113,8 @@
   "/regions/{region_id}": {
     "$ref": "./paths/Regions-region_id.json"
   },
-  "/register-shipment": {
-    "$ref": "./paths/RegisterShipment.json"
+  "/registered-shipments": {
+    "$ref": "./paths/RegisteredShipments.json"
   },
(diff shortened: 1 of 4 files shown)


specification/paths.json
```diff
@@ -113,6 +113,9 @@
   "/regions/{region_id}": {
     "$ref": "./paths/Regions-region_id.json"
   },
+  "/register-shipment": {
```
NickVries, line 116: I'm not convinced about this endpoint name. I believe it may cause confusion to users, making them think they need to call this endpoint in order to register any shipment with the carrier.
Maybe `create-and-register-shipment`?

NickVries (changes requested):

specification/paths.json
M4tini: I created a hidden `POST /registered-shipments` endpoint with `application/vnd.api+json` as discussed.

NickVries (approved):

PR #522, opened 2020-12-01
Added meta with file contents to registered-shipments response

yoan-myparcel:
# Changes
Added a meta array of files (including contents) to the response of POST /registered-shipments

### DIFF
--- a/specification/paths/RegisteredShipments.json
+++ b/specification/paths/RegisteredShipments.json
@@ -80,6 +80,37 @@
                       }
                     }
                   ]
+                },
+                "meta": {
+                  "type": "object",
+                  "properties": {
+                    "files": {
+                      "type": "array",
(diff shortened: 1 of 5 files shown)


specification/paths/RegisteredShipments.json
```diff
@@ -80,6 +80,37 @@
...
+                        "additionalProperties": false,
+                        "properties": {
+                          "extension": {
+                            "type": "string",
+                            "example": "pdf"
+                          },
+                          "mime_type": {
+                            "type": "string",
+                            "example": "application/pdf"
+                          },
```
M4tini, line 106: Let's make a `FileFormat.json` and use that in the `File.json` and here (with `allOf` "contents").

M4tini (changes requested):
Also, for international shipments it will not be clear which of the three PDF files in the meta is the `label`...

M4tini (approved):

PR #537, opened 2021-01-20
✨ Tax and duty in shipment response

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3859
(also removed unused ShipmentItem schema)


specification/schemas/BaseCustoms.json
```diff
@@ -38,19 +38,14 @@
...
+          "$ref": "#/components/schemas/Price"
         },
         {
-          "type": "null"
+          "type": "null",
+          "description": ""
         }
-      ]
+      ],
+      "description": "The shipping cost paid by the customer, for tax and duty purposes."
```
yoan-myparcel, line 48: Why not the description up there in the previous field that gets merged?

yoan-myparcel (approved):

specification/schemas/BaseCustoms.json
M4tini: Because then Redoc does not show it when the item is collapsed. This is the only way I got it to show nicely, but it involved having an empty description for the null, other wise it would show twice 💩

PR #540, opened 2021-01-27
🚀

M4tini:


specification/schemas/BaseShipmentItem.json
```diff
@@ -63,6 +63,30 @@
           "type": "null"
         }
       ]
+    },
+    "item_tax": {
```
yoan-myparcel, line 67: This we change to `tax`, right?

specification/schemas/BaseShipmentItem.json
```diff
@@ -63,6 +63,30 @@
...
+        {
+          "readOnly": true,
+          "type": "null"
+        },
+        {
+          "$ref": "#/components/schemas/Price"
+        }
+      ]
+    },
+    "item_duty": {
```
yoan-myparcel, line 79: and this to `duty`?

PR #545, opened 2021-02-25
MP-4038 ✨ Introduce shipments.view scope

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3908
I decided to give `shipments.view` on GET shipments + all GET endpoints to retrieve Shipment relationship resources.


yoan-myparcel:
@M4tini I assume that shipments.view also implies reports. Lets put the scope there as well!

yoan-myparcel (approved):

PR #560, opened 2021-04-17
✨ Introduced new scope

yoan-myparcel:
# Changes
Experimental scope will not be enough. We want to expose logs information only to specific users.

### DIFF
--- a/specification/paths/ExportLoginAttempts.json
+++ b/specification/paths/ExportLoginAttempts.json
@@ -6,7 +6,8 @@
     "security": [
       {
         "OAuth2": [
-          "experimental"
+          "experimental",
+          "logs.view"
         ]
       }
     ],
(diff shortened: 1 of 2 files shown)


specification/paths/ExportLoginAttempts.json
```diff
@@ -6,7 +6,8 @@
     "security": [
       {
         "OAuth2": [
-          "experimental"
+          "experimental",
+          "logs.view"
```
M4tini, line 10: - Let's list this scope on top of the spec with the other scopes.
- Do we still need the experimental scope on this endpoint?

M4tini (changes requested):

specification/paths/ExportLoginAttempts.json
yoan-myparcel: I suggest we keep the experimental scope so that it is only used by the portal.

yoan-myparcel: On second thought - no, we do not really need it. I will remove it

M4tini (approved):

PR #567, opened 2021-05-10
MP-4229 ✨ Add meta with collo details to multi-colli response

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-4170
- Fixed bug where `colli` relationship contained entire shipment resources instead of `id` + `type`.
- Added `meta` with base64 filedata to `colli` relationship of `/register-multi-colli-shipment` response.
- Fixed authentication token URL.

### DIFF
--- a/specification/schemas/ShipmentColliRelationship.json
+++ b/specification/schemas/ShipmentColliRelationship.json
@@ -8,7 +8,7 @@
     "data": {
       "type": "array",
       "items": {
-        "$ref": "#/components/schemas/Shipment"
+        "$ref": "#/components/schemas/ShipmentResource"
       }
     }
   }
--- a/root.json
+++ b/root.json
@@ -27,7 +27,7 @@
         "scheme": "Bearer",
         "flows": {
           "clientCredentials": {
-            "tokenUrl": "$AUTH_HOST",
+            "tokenUrl": "$AUTH_HOST/access-token",
             "scopes": {
               "addresses.suggest": "Retrieve address suggestions.",
               "broker.member": "Retrieve broker related resources and create organizations for that broker.",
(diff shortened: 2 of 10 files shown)


specification/schemas/ShipmentColliRelationship.json
```diff
@@ -8,7 +8,7 @@
     "data": {
       "type": "array",
       "items": {
-        "$ref": "#/components/schemas/Shipment"
+        "$ref": "#/components/schemas/ShipmentResource"
```
zwaans, line 11: I would expect a filename change in here as well then. Also, I'm not really sure about the 'resource' suffix since no other components use this.

Please elaborate if I'm missing something (:

zwaans (commented):
I'm not sure if my comment actually requests changes, just an conversation about the topic  maybe.

specification/schemas/ShipmentColliRelationship.json
M4tini: Good remark, and yes there is a difference between some resources in our api-specification.

Half of the resources have been untouched since we moved from Swagger 2.0 to 3.0 (OpenAPI). This update introduced `oneOf` and `allOf` which allowed us to combine partial resources. Then we came up with a more re-usable way of working: separate attribute requirements from resources. For example:
```json
{
  "allOf": [
    { "$ref": "#/components/schemas/Shipment" },
    { "required": ["recipient_address", "sender_address"] }
  ]
}
```
This new approach is explained in the PR that introduced it for three resources: https://github.com/MyParcelCOM/api-specification/pull/503
So far, we have reworked these resources: `Report` + `Shipment` + `Shop` + `Status` + `SystemMessage` + `User`.

zwaans (approved):

PR #583, opened 2021-06-18
✨ Introduced new tax identification number types vat and ioss

yoan-myparcel:
# Changes
- Introduced `vat` and `ioss` types


M4tini (approved):
I thought about adding `voec` but that can just be `vat` for country_code `NO`.

PR #593, opened 2021-07-07
MP-3746 🔩 Added the new endpoint to the api-spec

zwaans:
## Changes 
- Added the new broker - carrier relationship endpoint to the api-spec

### Links
- https://myparcelcombv.atlassian.net/browse/MP-3746

### DIFF
--- a/specification/paths/Brokers-broker_id-carriers.json
+++ b/specification/paths/Brokers-broker_id-carriers.json
@@ -1,0 +1,109 @@
+{
+  "parameters": [
+    {
+      "$ref": "#/components/parameters/path-broker_id"
+    }
+  ],
+  "get": {
+    "tags": [
+      "Brokers"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "enterprise.manage"
+        ]
+      }
+    ],
+    "summary": "Get the available carriers for a specific broker.",
+    "description": "This endpoint retrieves available carriers for a broker with the supplied id.",
+    "responses": {
+      "200": {
+        "description": "Retrieved the available carriers for the broker.",
+        "content": {
+          "application/vnd.api+json": {
+            "schema": {
+              "type": "object",
+              "required": [
+                "data"
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "data": {
+                  "type": "array",
+                  "items": {
+                    "allOf": [
+                      {
+                        "$ref": "#/components/schemas/Carrier"
+                      },
+                      {
+                        "required": [
+                          "id",
+                          "links"
+                        ],
+                        "properties": {
+                          "attributes": {
+                            "required": [
+                              "status",
+                              "currency",
+                              "name",
+                              "created_at"
+                            ]
+                          },
+                          "relationships": {
+                            "properties": {
+                              "owner": {
+                                "example": {
+                                  "links": {
+                                    "related": "$API_HOST/carriers/9cdf86e8-333f-4ed9-bb31-4935c780c947"
+                                  }
+                                }
+                              }
+                            }
+                          }
+                        }
+                      }
+                    ]
+                  }
+                },
+                "meta": {
+                  "$ref": "#/components/schemas/PaginationMeta"
+                },
+                "links": {
+                  "type": "object",
+                  "required": [
+                    "self",
+                    "first",
+                    "last"
+                  ],
+                  "properties": {
+                    "self": {
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=3&page[size]=30"
+                    },
+                    "first": {
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=1&page[size]=30"
+                    },
+                    "prev": {
+                      "type": "string",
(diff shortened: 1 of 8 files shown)


specification/paths/Brokers-broker_id-carriers.json
```diff
@@ -0,0 +1,109 @@
...
+    }
+  ],
+  "get": {
+    "tags": [
+      "Brokers"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "enterprise.manage"
```
M4tini, line 14: Since it's a GET endpoint, should we allow `broker.member` to access it as well?

specification/paths/Brokers-broker_id-carriers.json
```diff
@@ -0,0 +1,109 @@
...
+                  "type": "object",
+                  "required": [
+                    "self",
+                    "first",
+                    "last"
+                  ],
+                  "properties": {
+                    "self": {
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=3&page[size]=30"
```
M4tini, line 82: Can you please do a global search on `API_HOST` and make sure all of them are `$API_HOST`. I noticed the `$` is missing in multiple (old) files.

specification/paths/Brokers-broker_id-carriers.json
```diff
@@ -0,0 +1,109 @@
...
+                            "required": [
+                              "status",
+                              "currency",
+                              "name",
+                              "created_at"
+                            ]
+                          },
+                          "relationships": {
+                            "properties": {
+                              "owner": {
```
M4tini, line 55: A carrier has no relationships.

specification/paths/Brokers-broker_id-carriers.json
```diff
@@ -0,0 +1,109 @@
...
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=2&page[size]=30"
+                    },
+                    "next": {
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=4&page[size]=30"
+                    },
+                    "last": {
+                      "type": "string",
+                      "example": "API_HOST/organizations/9cdf86e8-333f-4ed9-bb31-4935c780c947/contracts?page[number]=13&page[size]=30"
```
M4tini, line 98: All the links above should be replaced with `/brokers/a294ee55-bc94-4890-b734-afd56c158f95/carriers`

specification/paths/Brokers-broker_id-relationships-carriers.json
```diff
@@ -0,0 +1,67 @@
...
+    ],
+    "summary": "Add carrier relationships to a broker.",
+    "description": "This endpoint adds carrier relationships to the broker with the supplied id.",
+    "requestBody": {
+      "description": "The carrier relationships to be added to the broker.",
+      "required": true,
+      "content": {
+        "application/vnd.api+json": {
+          "schema": {
+            "$ref": "#/components/schemas/CarrierRelationship"
```
M4tini, line 26: We should support adding multiple carriers at once, so let's make a `CarriersRelationship` component schema (for it to be recognized, the `CarriersRelationship` should also be mentioned in `specification/schemas.json`)

specification/paths/Brokers-broker_id-relationships-carriers.json
```diff
@@ -0,0 +1,67 @@
...
+    ],
+    "summary": "Delete carrier relationships from a broker.",
+    "description": "This endpoint deletes carrier relationships from the broker with the supplied id.",
+    "requestBody": {
+      "description": "The carrier relationships to be deleted from the broker.",
+      "required": true,
+      "content": {
+        "application/vnd.api+json": {
+          "schema": {
+            "$ref": "#/components/schemas/CarrierRelationship"
```
M4tini, line 56: `CarriersRelationship`

M4tini (changes requested):

specification/paths/Brokers-broker_id-carriers.json
zwaans: I'm having a hard time knowing which scope is used in what scenario. If users with the `broker.member` scope have needs for viewing which carriers the broker uses then we should add it.

specification/paths/Brokers-broker_id-carriers.json
zwaans: Ah so these 'examples' reference actual existing entities?

zwaans: oh now I see I haven't renamed the example organizations into brokers. My bad.

specification/paths/Brokers-broker_id-carriers.json
M4tini: We can also wait until the need is there. It's only used by admins for now.

specification/paths/Brokers-broker_id-carriers.json
M4tini: Yeah that was the main thing :) The UUID values are not referencing anything, but we aim to use the same UUID when we talk about the same resource (and have each resource have a different one).

M4tini (approved):

PR #600, opened 2021-08-12
🛬 MP-4385 Added hscode suggestion rpc endpoint to api spec

zwaans:
# Changes
- Added hs code suggestion endpoint to the api spec

## Related issue
- https://myparcelcombv.atlassian.net/browse/MP-4385

### DIFF
--- a/specification/paths/Suggest-hscode.json
+++ b/specification/paths/Suggest-hscode.json
@@ -61,30 +61,30 @@
                   "items": {
                     "type": "object",
                     "required": [
-                      "hurricaneUUID",
-                      "hsCode",
-                      "commodityCode",
-                      "vectorWeight",
-                      "productDescription"
+                      "external_id",
+                      "hs_code",
+                      "commodity_code",
+                      "vector_weight",
+                      "product_description"
                     ],
                     "properties": {
-                      "hurricaneUUID": {
+                      "external_id": {
                         "type": "string",
                         "example": "13822608-3a7e-494f-8bbf-c89245641bbe"
                       },
-                      "hsCode": {
+                      "hs_code": {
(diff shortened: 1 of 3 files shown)


specification/paths/Suggest-hscode.json
```diff
@@ -0,0 +1,104 @@
...
+          "schema": {
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "properties": {
+              "data": {
+                "required": [
+                  "clientSKU",
+                  "clientDescription"
```
yoan-myparcel, line 29: I saw this in the landed cost PR but I thought it was for fields we send to hurricane directly. I don't think it's good that we use snake case in the entire spec and camel case here. Also the "client" prefix is somewhat  unnecessary. Same goes for "source". We should really stick to field names that are used in other endpoints too (recipient instead of client?). 

I assume that this is also a problem affecting the frontend code and not only the spec and landed cost. Lets fix it now so it is not too late.

specification/paths/Suggest-hscode.json
```diff
@@ -0,0 +1,104 @@
...
+                "data": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "required": [
+                      "hurricaneUUID",
+                      "hsCode",
+                      "commodityCode",
+                      "vectorWeight",
+                      "productDescription"
```
yoan-myparcel, line 71: Again, here we should stick as close as possible to field names we already have in the specs. These values will be used when creating shipments so field names should be available to choose from.

yoan-myparcel (dismissed):

M4tini (approved):

PR #602, opened 2021-08-13
🛬 MP-4449 Removed unused shipment fields from request

zwaans:
# Changes
- Removed fields that are not needed to make the landed cost calculation
- Flattened the request array

## Related issue
- https://myparcelcombv.atlassian.net/browse/MP-4449


specification/paths/CalculateLandedCosts.json
```diff
@@ -87,7 +89,19 @@
               ],
               "properties": {
                 "data": {
-                  "$ref": "#/components/schemas/Shipment"
+                  "type": "object",
+                  "required": [
+                    "total_tax",
+                    "total_duty"
+                  ],
+                  "properties": {
```
M4tini, line 97: We also output the items with individual `tax` and `duty` and actually also the `hs_code` if we needed to look it up.

specification/paths/CalculateLandedCosts.json
```diff
@@ -35,26 +35,28 @@
                     "properties": {
                       "attributes": {
                         "required": [
-                          "recipient_address",
-                          "sender_address",
-                          "physical_properties",
```
M4tini, line 40: I think we should remove the `allOf` with the `Shipment` component and make this a separate data object.

M4tini (changes requested):

specification/paths/CalculateLandedCosts.json
zwaans: I think we should then rediscuss this on monday (i'm a bit gaar atm). Because that doesnt add up with the output we discussed earlier: ```data: {
  total_tax: {
    amount:
    currency:
  },
  total_duty: {
    amount:
    currency:
  }
}```

M4tini (approved):

M4tini (approved):

PR #603, opened 2021-08-16
MP-4454 Address endpoints

NickVries:
## Changes
- Changed Organization and PickupDropoffLocation schema's to use BaseAddress schema instead of Address schema.
- Introduced AddressResource schema for `type` and `id` properties of the Address resource.
- Changed Address schema to be a full resource schema.
- Introduced the following endpoints:
  - GET `/addresses`
  - GET `/addresses/{address_id}`
  - POST `/addresses`
  - PATCH `/addresses/{address_id}`
  - DELETE `/addresses/{address_id}`

## Related Issues
- [Subtask] https://myparcelcombv.atlassian.net/browse/MP-4454
- [Story] https://myparcelcombv.atlassian.net/browse/MP-4417


specification/paths/Addresses-address_id.json
```diff
@@ -0,0 +1,151 @@
...
+          "organizations.manage",
+          "shipments.manage"
+        ]
+      }
+    ],
+    "summary": "Get an address by id.",
+    "description": "This endpoint retrieves the address with the supplied id.",
+    "responses": {
+      "200": {
+        "description": "Retrieved the shop.",
```
M4tini, line 23:
```suggestion
        "description": "Retrieved the address.",
```

specification/paths/Addresses.json
```diff
@@ -0,0 +1,181 @@
...
+          "type": "string"
+        }
+      },
+      {
+        "name": "filter[organization]",
+        "in": "query",
+        "description": "Comma separated string of organization ids to filter by.",
+        "schema": {
+          "type": "string"
+        }
```
M4tini, line 31: This would be nice (also on other places), but we should rename that file because it's not shipment specific.
```suggestion
        "$ref": "#/components/parameters/query-shipments-filter-organization"
```

M4tini (changes requested):
Please update `tags.json` with an Addresses entry (to sort the menu).

M4tini (approved):

PR #613, opened 2021-10-01
🦺 MP 4541 api-spec service options values

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-4541

## Changes
- Updated the service option ids on the multicolli endpoint to hold objects for service option values instead of just an array of service option ids.

### DIFF
--- a/specification/paths/RegisterMultiColliShipment.json
+++ b/specification/paths/RegisterMultiColliShipment.json
@@ -142,7 +142,31 @@
                   "service_option_ids": {
                     "type": "array",
                     "items": {
-                      "$ref": "#/components/schemas/Uuid"
+                      "type": "object",
+                      "required": [
+                        "id",
+                        "values"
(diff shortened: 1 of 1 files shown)


specification/paths/RegisterMultiColliShipment.json
```diff
@@ -142,7 +142,31 @@
                   "service_option_ids": {
                     "type": "array",
                     "items": {
-                      "$ref": "#/components/schemas/Uuid"
+                      "type": "object",
+                      "required": [
+                        "id",
+                        "values"
```
yoan-myparcel, line 148: This is actually not required! Most service options do not have values_format!
```suggestion
                        "id"
```

yoan-myparcel (changes requested):

yoan-myparcel (approved):

PR #633, opened 2021-11-03
♻️ Allow broker members to get enterprises

M4tini:
Brokers have an Enterprise relationship, so the links in there should be retrievable with the same token. Since the broker endpoints have `broker.member` and `enterprises.manage` scopes, the enterprise endpoints should have this as well.

### DIFF
--- a/specification/paths/Brokers-broker_id-relationships-carriers.json
+++ b/specification/paths/Brokers-broker_id-relationships-carriers.json
@@ -11,6 +11,7 @@
     "security": [
       {
         "OAuth2": [
+          "broker.member",
           "enterprise.manage"
         ]
       }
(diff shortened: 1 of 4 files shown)


specification/paths/Brokers-broker_id-relationships-carriers.json
```diff
@@ -11,6 +11,7 @@
     "security": [
       {
         "OAuth2": [
+          "broker.member",
```
NickVries, line 14: Why are we adding the broker.member scope to the POST carrier relationship endpoint?

NickVries (changes requested):

specification/paths/Brokers-broker_id-relationships-carriers.json
M4tini: Whoops, good catch! This was supposed to be set on the `GET /brokers/{broker}/carriers` endpoint.

NickVries (approved):

PR #646, opened 2021-12-09
Introduced new RPC endpoint /count-report-entries

yoan-myparcel:
# Changes
- Added new RPC endpoint for counting report entries
- Introduced BaseReport schema that does not have the "name" attribute

### DIFF
--- a/specification/schemas/Report.json
+++ b/specification/schemas/Report.json
@@ -1,7 +1,7 @@
 {
   "allOf": [
     {
-      "$ref": "#/components/schemas/ReportResource"
+      "$ref": "#/components/schemas/BaseReport"
     },
     {
       "properties": {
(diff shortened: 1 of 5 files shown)


specification/schemas/Report.json
```diff
@@ -1,149 +1,17 @@
...
-                    "country_code": {
-                      "$ref": "#/components/schemas/CountryCode"
-                    }
-                  }
-                }
-              }
-            },
             "name": {
               "type": "string",
               "example": "2020 Q1 UK shipment report"
```
M4tini, line 17: I think the count endpoint should not crash when a name is sent with the report, because it is a valid property on a `Report` resource. Let's move the `name` to the `BaseReport` file as well and keep it optional.

M4tini (changes requested):

specification/schemas/Report.json
yoan-myparcel: I see, making the field optional will fix the issue indeed.

yoan-myparcel: @M4tini fixed!

M4tini (approved):

PR #670, opened 2022-01-28
✨  Import resource action on hooks

NickVries:
## Changes
- Added `import` as possible `resource_action` on hooks.
- Removed `delete` as possible `resource_action` on hooks since it's not used and our docs also doesn't mention it.
- Added filter[trigger][resource_action] filter to the hooks endpoint. We already have filter[action_type]. Should we rename that to filter[action][action_type] :question: 

## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-4812 - Subtask
- https://myparcelcombv.atlassian.net/browse/MP-4575 - Story

### DIFF
--- a/specification/paths/Hooks.json
+++ b/specification/paths/Hooks.json
@@ -30,6 +30,19 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
+        "name": "filter[trigger][resource_action]",
+        "in": "query",
+        "description": "[Resource action](https://docs.myparcel.com/api/resources/hooks/trigger/) to filter the hooks by.",
+        "schema": {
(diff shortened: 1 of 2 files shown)


zwaans (commented):
Good point regarding the naming! I like this consistency. If we rename filter[action_type] to filter[action][action_type] in order to match with the new filter[trigger][resource_action] shouldn't the new trigger be called `filter[trigger][trigger_type]`?

NickVries:
As discussed during the standup, I will deprecate/remove the `filter[action_type]` filter and add `filter[action][action_type]`. We should still support the old filter in the backend to avoid breaking changes.

PR #679, opened 2022-03-10
⚡MP-4895 Added collection meta object

zwaans:
# Changes
- Expanded the registered shipments object with a collection object in the meta attribute.

## Related issue
- https://myparcelcombv.atlassian.net/browse/MP-4895

### DIFF
--- a/specification/paths/RegisteredShipments.json
+++ b/specification/paths/RegisteredShipments.json
@@ -38,18 +38,18 @@
                       },
                       "collection_date": {
                         "type": "string",
-                        "format": "date-time"
+                        "example": "2022-04-04"
                       },
                       "collection_time": {
                         "type": "object",
                         "properties": {
                           "from": {
                             "type": "string",
-                            "format": "time"
+                            "example": "12:00"
                           },
                           "to": {
                             "type": "string",
-                            "format": "time"
+                            "format": "16:00"
                           }
                         }
                       }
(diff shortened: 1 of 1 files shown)


specification/paths/RegisteredShipments.json
```diff
@@ -30,6 +30,31 @@
...
                 "properties": {
+                  "collection": {
+                    "type": "object",
+                    "properties": {
+                      "name": {
+                        "type": "string"
+                      },
+                      "collection_date": {
+                        "type": "string",
+                        "format": "date-time"
```
M4tini, line 41: I think this should be `date` (if that format means Y-m-d) because the time is specified in `collection_time`.

M4tini (changes requested):
I assume we did not list `address` as part of the collection, because we will take the `sender_address` of the shipment. Should we list it as optional parameter or leave it for now? (we can always add it later)

zwaans:
> I assume we did not list `address` as part of the collection, because we will take the `sender_address` of the shipment. Should we list it as optional parameter or leave it for now? (we can always add it later)

Is that really a usecase that we want to support? instead of it being collected from a shop please come and pick it up at this other location? If so then it would make sense to add it. Maybe @yoan-myparcel can weigh in on this.

M4tini (approved):

PR #687, opened 2022-03-24
MP-4883 Collections endpoints

NickVries:
## Changes
- Added schemas for Collection resources.
- Added the following endpoints:
  - GET /collections
  - GET /collections/{collection_id}
  - POST /collections
  - PATCH /collections/{collection_id}
  - DELETE /collections/{collection_id}

## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-4883

## ⚠️ Note ⚠️
I think we should wait with merging this PR until the API work has been done/started. 
Otherwise we'll end up with endpoints in our specification that do not yet exist in our API.

### DIFF
--- a/specification/paths/Collections.json
+++ b/specification/paths/Collections.json
@@ -156,6 +156,12 @@
                             ]
                           }
                         }
+                      },
+                      "relationships": {
+                        "required": [
+                          "contract",
+                          "shop"
+                        ]
                       }
                     }
                   }
--- a/specification/paths/Collections.json
+++ b/specification/paths/Collections.json
@@ -1,0 +1,191 @@
+{
+  "get": {
+    "tags": [
+      "Collections"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "shipments.manage",
+          "shipments.view"
+        ]
+      }
+    ],
+    "summary": "Get all available collections.",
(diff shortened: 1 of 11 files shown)


specification/paths/Collections.json
```diff
@@ -0,0 +1,197 @@
...
+                          "attributes": {
+                            "properties": {
+                              "tracking_code": {
+                                "type": "string",
+                                "example": "TRC123456"
+                              }
+                            }
+                          }
+                        }
+                      }
```
M4tini, line 57: I would prefer this attribute to be part of the `Collection` schema (read-only) and indicate the presence of the value in the description. Because now we have these 12 lines duplicated in 3 different places + if I would search for a complete list of Collection resource properties, I would check the collection.json schema and not expect additional properties in response paths.

I see why you want to do it this way though, but our specification is structured in another way. There are many attributes and especially relationships which are not present or even able to have a value in a POST response. They are just possible properties of the particular **resource** being created, so all endpoints simply describe that resource and the properties that definitely have a value.

Let's check with the team tomorrow. If we want to be this explicit, we should modify other schemas as well.

M4tini (changes requested):

M4tini (approved):

PR #688, opened 2022-04-11
🔍 Introduce category on Carrier status

zwaans:
# Changes
- Introduced a new enum for carrier exceptional statuses based on the list of statuses supplied in https://myparcelcombv.atlassian.net/browse/MP-4150


specification/schemas/CarrierExceptionalStatus.json
```diff
@@ -0,0 +1,24 @@
...
+    "failed-shipment-missing",
+    "failed-shipment-damaged",
+    "delay",
+    "relabelling",
+    "partial-collection",
+    "partial-delivery",
+    "change-in-flight"
+  ],
+  "example": "failed-delivery-attempt",
+  "default": "failed-delivery-attempt",
```
M4tini, line 22: This should have a different or no default value. Was it not called `default`? Should we add it to the enum?

M4tini (changes requested):

specification/schemas/CarrierExceptionalStatus.json
zwaans: @M4tini I was about to bring this up during stand up haha

PR #689, opened 2022-04-13
Added status category filter

zwaans:
# Changes
- Introduced a new filter for status categories on shipments

### DIFF
--- a/specification/parameters/query-shipments-filter-carrier_status-category.json
+++ b/specification/parameters/query-shipments-filter-carrier_status-category.json
@@ -1,7 +1,7 @@
 {
   "name": "filter[carrier_status][category]",
   "in": "query",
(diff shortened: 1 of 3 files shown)


specification/parameters/query-shipments-filter-carrier_status-category.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "filter[carrier_status][category]",
+  "in": "query",
+  "description": "Comma separated string of status categories to filter by. Only shipments with a status that corresponds with one of the categories will be in the response.",
```
NickVries, line 4: > Only shipments with a status that corresponds with one of the categories will be in the response.

Let's clarify here what "shipments with a status" means. Does it only check the current shipment status, or will it also return shipments that have that category in any of their historical shipment statuses?

NickVries (changes requested):

NickVries (approved):

PR #696, opened 2022-04-28
patchable collections

zwaans:
# Changes
- Added register attribute on collections

# Notes
- Somehow the preview request on the right shows the register property on the patch endpoint but the main pane in the middle doesn't.

### DIFF
--- a/specification/paths/Collections-collection_id.json
+++ b/specification/paths/Collections-collection_id.json
@@ -94,7 +94,6 @@
                             "type": "boolean",
                             "example": true,
                             "description": "Indicates whether the collection should be registered with the carrier. Further updates are no longer possible after a successful registration.",
-                            "readOnly": false
                           },
                           "address": {
                             "readOnly": true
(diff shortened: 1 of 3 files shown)


specification/schemas/Collection.json
```diff
@@ -16,6 +16,12 @@
             "address": {
               "$ref": "#/components/schemas/BaseAddress"
             },
+            "register": {
+              "type": "boolean",
+              "example": true,
+              "description": "Indicates whether the collection should be registered with the carrier. Further updates are no longer possible after a successful registration.",
+              "readOnly": true
```
M4tini, line 23: Let's not define readOnly on the resource, but only in response paths.

specification/paths/Collections-collection_id.json
```diff
@@ -93,7 +93,8 @@
                           "register": {
                             "type": "boolean",
                             "example": true,
-                            "description": "Indicates whether the collection should be registered with the carrier. Further updates are no longer possible after a successful registration."
+                            "description": "Indicates whether the collection should be registered with the carrier. Further updates are no longer possible after a successful registration.",
+                            "readOnly": false
```
M4tini, line 97: Let's revert this and instead add `"readOnly": true` to the request body of the `post` path

M4tini (changes requested):

M4tini (approved):

PR #705, opened 2022-05-09
🥅 Get collections filter params

zwaans:
# changes
- introduced filterable params for the get collections endpoint


specification/paths/Collections-collection_id.json
```diff
@@ -26,6 +26,38 @@
...
+          "type": "string"
+        }
+      },
+      {
+        "name": "filter[shop]",
+        "in": "query",
+        "description": "Comma separated string of shop ids to filter by.",
+        "schema": {
+          "type": "string"
+        }
```
M4tini, line 44:
```suggestion
        "$ref": "#/components/parameters/query-filter-shop"
```
FYI - there is already a `query-shipments-filter-shop` which we should rename (and also the references to it).

specification/paths/Collections-collection_id.json
```diff
@@ -26,6 +26,38 @@
...
+      },
+      {
+        "name": "filter[carrier]",
+        "in": "query",
+        "description": "Comma separated string of carrier ids to filter by.",
+        "schema": {
+          "type": "string"
+        }
```
M4tini, line 60: Same story for `query-shipments-filter-carrier`

M4tini (changes requested):
Maybe we want to make `filter[status]` like this as well, because it's quite generic.

yoan-myparcel (approved):

PR #710, opened 2022-06-02
♻️ Restructure multi-colli POST endpoint

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-5041
---
- Deprecate multi-colli RPC endpoint.
- Create POST /multi-colli-shipments endpoint with colli in meta.

### DIFF
--- a/specification/paths/MultiColliShipments.json
+++ b/specification/paths/MultiColliShipments.json
@@ -1,0 +1,281 @@
+{
+  "post": {
+    "tags": [
+      "Shipments"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "shipments.manage"
+        ]
+      }
+    ],
+    "summary": "Create and register a multi-colli shipment.",
+    "description": "This endpoint creates a new shipment consisting of multiple parcels to the same address, which is then immediately communicated to the carrier. The carrier response is processed before your request is completed, so files will be available.",
+    "parameters": [
+      {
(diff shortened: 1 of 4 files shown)


specification/paths/MultiColliShipments.json
```diff
@@ -0,0 +1,281 @@
...
+                        "links"
+                      ],
+                      "properties": {
+                        "relationships": {
+                          "required": [
+                            "contract",
+                            "files",
+                            "service",
+                            "shipment_status",
+                            "shop"
```
yoan-myparcel, line 233: We need the colli relationship too, with meta data for the file contents

specification/paths/MultiColliShipments.json
```diff
@@ -0,0 +1,281 @@
...
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                },
+                "meta": {
```
yoan-myparcel, line 240: This might not be needed since the contents of the labels is in the colli relationship

yoan-myparcel (changes requested):

specification/paths/MultiColliShipments.json
M4tini: I replaced `Shipment` with `ShipmentResponse` which already contains `colli` and some other requirements. Not sure why the `/registered-shipments` endpoint did not use `ShipmentResponse` in it's response, so I changed it there as well.

yoan-myparcel (approved):

PR #723, opened 2022-07-14
MP-5099 Added service rate import schema.

NickVries:
## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-5099

### DIFF
--- a/specification/schemas.json
+++ b/specification/schemas.json
@@ -332,6 +332,9 @@
   "ServiceRate": {
     "$ref": "./schemas/ServiceRate.json"
   },
+  "ServiceRateImport": {
+    "$ref": "./schemas/ServiceRateImport.json"
+  },
   "ServiceRateIncludes": {
     "$ref": "./schemas/ServiceRateIncludes.json"
   },
--- a/specification/schemas.json
+++ b/specification/schemas.json
@@ -332,8 +332,8 @@
   "ServiceRate": {
     "$ref": "./schemas/ServiceRate.json"
   },
-  "ServiceRateImport": {
-    "$ref": "./schemas/ServiceRateImport.json"
+  "RatesImport": {
+    "$ref": "./schemas/RatesImport.json"
   },
   "ServiceRateIncludes": {
     "$ref": "./schemas/ServiceRateIncludes.json"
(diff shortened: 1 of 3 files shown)


specification/schemas.json
```diff
@@ -332,6 +332,9 @@
   "ServiceRate": {
     "$ref": "./schemas/ServiceRate.json"
   },
+  "ServiceRateImport": {
+    "$ref": "./schemas/ServiceRateImport.json"
```
yoan-myparcel, line 336:
```suggestion
    "$ref": "./schemas/RateImport.json"
```

specification/schemas/ServiceRateImport.json
```diff
@@ -0,0 +1,64 @@
+{
+  "type": "object",
+  "required": [
+    "type"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
+      "pattern": "^service-rate-imports$",
+      "example": "service-rate-imports"
```
yoan-myparcel, line 10:
```suggestion
      "pattern": "^rate-imports$",
      "example": "rate-imports"
```

specification/schemas.json
```diff
@@ -332,6 +332,9 @@
   "ServiceRate": {
     "$ref": "./schemas/ServiceRate.json"
   },
+  "ServiceRateImport": {
```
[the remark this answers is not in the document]
NickVries: The resource that it imports are ServiceRates though. Even the prices of the options are stored on the ServiceRate resource. Wouldn't it be more clear to know that ServiceRateImport results in ServiceRates instead of RateImport retulting in ServiceRates?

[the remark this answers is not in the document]
yoan-myparcel: I think of this schema as somewhat isolated from the service rates actually. The resource itself as represented in the api specification has no relation to service rates, it wrap an importable file, hence I am suggesting a simpler name for the users. Essentially, people know what rates are. The question "how can I import my rates" has a simple answer "through rates-import api". 

I just do not see practically how having "Service" as a prefix makes it more clear, I personally prefer a shorter, simpler and direct name.

specification/schemas/ServiceRateImport.json
yoan-myparcel: Here the plural on imports is ok because I think `rates-imports` is not exactly gramatically correct.

yoan-myparcel (approved):

PR #736, opened 2022-08-01
[api-spec] :office: :lock: MP-5093 added new type indication and filter for location type

zwaans:


specification/schemas/PickupDropoffLocation.json
```diff
@@ -49,6 +50,14 @@
                 }
               ]
             },
+            "location_type":{
+              "type": "string",
+              "default": "Office",
```
M4tini, line 55:
```suggestion
              "default": "office",
```

M4tini (approved):

PR #739, opened 2022-08-22
RatesImport Added missing file relationship

zwaans:


M4tini (changes requested):
Why do we expose the file? I though this was only for internal use, to be able to process the contents in a scheduled job.

specification/schemas/RatesImport.json
```diff
@@ -57,6 +57,16 @@
       "properties": {
         "contract": {
           "$ref": "#/components/schemas/ContractRelationship"
+        },
+        "files": {
+          "allOf": [
+            {
+              "$ref": "#/components/schemas/FilesRelationship"
```
M4tini, line 64:
```suggestion
        "file": {
          "allOf": [
            {
              "$ref": "#/components/schemas/FileRelationship"
```

M4tini (changes requested):

M4tini (approved):

PR #754, opened 2022-10-03
♻️ Remove regex from zone country_codes to improve error messages

M4tini:
We will add API validation to be able to output which country code is invalid.


NickVries (approved):

NickVries:
I've approved it, but won't merge until we have actual validation in the API.

PR #772, opened 2022-11-10
♻️ Require users.manage and return user on post

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-5348
---
- All user endpoints should require `users.manage`.
- `POST /users` should return a resource instead of an empty response.

### DIFF
--- a/specification/paths/Users-user_id.json
+++ b/specification/paths/Users-user_id.json
@@ -11,6 +11,7 @@
     "security": [
       {
         "OAuth2": [
+          "users.manage"
         ]
       }
     ],
@@ -45,6 +46,7 @@
     "security": [
       {
         "OAuth2": [
+          "users.manage"
         ]
       }
     ],
--- a/specification/paths/Users-user_id.json
+++ b/specification/paths/Users-user_id.json
@@ -11,12 +11,11 @@
     "security": [
       {
         "OAuth2": [
-          "users.manage"
         ]
       }
     ],
     "summary": "Get a user by id.",
-    "description": "This endpoint retrieves the user with the supplied id.",
+    "description": "This endpoint retrieves the user with the supplied id. You can always retrieve your own user, but you need the `users.manage` scope to retrieve other users.",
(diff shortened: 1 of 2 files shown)


specification/schemas/User.json
```diff
@@ -9,6 +9,10 @@
           "type": "object",
           "additionalProperties": false,
           "properties": {
+            "username": {
+              "type": "string",
+              "example": "s.holmes@holmesinvestigations.com"
+            },
```
M4tini, line 15:
```suggestion
```

specification/schemas/UserResponse.json
```diff
@@ -11,6 +11,7 @@
       "properties": {
         "attributes": {
           "required": [
+            "username",
```
NickVries, line 14:
```suggestion

```

specification/paths/Users.json
```diff
@@ -117,6 +124,7 @@
                       },
                       "attributes": {
                         "required": [
+                          "username",
```
NickVries, line 127:
```suggestion

```

NickVries (changes requested):

specification/paths/Users-user_id.json
```diff
@@ -45,6 +46,7 @@
     "security": [
       {
         "OAuth2": [
+          "users.manage"
```
M4tini, line 49: @NickVries I'm thinking: we do not give users `users.manage` on their own user.

We could require the scope, but another way is to let the `/users/{user_id}` endpoint check the `user_id` from the current token and allow `getOne` when it matches, so we (and the brokers) do not have to give `users.manage` on everyone.

M4tini: @NickVries This endpoint was already silently protected by `users.manage`. It was not listed here because you can also query this endpoint without the scope, fetching your own user.

I think we should remove this line and still not mention it here, what do you think?

NickVries (approved):

PR #778, opened 2022-11-24
:rotating_light: Fixed patch collection status relationship

zwaans:


specification/paths/Collections-collection_id.json
```diff
@@ -104,8 +104,17 @@
...
                         "properties": {
+                          "status": {
+                            "allOf": [
+                              {
+                                "$ref": "#/components/schemas/StatusRelationship"
+                              },
+                              {
+                                "description": "An untracked collection can be manually updated to indicate the collection has been collected."
+                              }
+                            ]
```
M4tini, line 116: This relationship is already a StatusRelationship, so the `allOf` can be replaced with just:
```
"description": "An untracked collection can be manually updated to indicate the collection has been collected."
```

M4tini (changes requested):

PR #784, opened 2022-12-15
MP-5488 Add registered shipments to a collection

zwaans:
# Changes
- Introduced new boolean property on the carrier to indicate if registered shipments can be added to a collection
- Introduced new rpc endpoint to bulk add shipments to a collection

### DIFF
--- a/specification/paths/AddShipmentsToCollection.json
+++ b/specification/paths/AddShipmentsToCollection.json
@@ -1,0 +1,57 @@
+{
+  "post": {
+    "tags": [
+      "RPC"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "shipments.manage",
+          "collections.manage"
+        ]
+      }
+    ],
+    "summary": "Add shipments to collection.",
+    "description": "This endpoint can add multiple shipments to a concept collection.",
+    "requestBody": {
+      "description": "The collection id and the shipment ids.",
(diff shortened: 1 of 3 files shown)


specification/paths/AddShipmentsToCollection.json
```diff
@@ -0,0 +1,57 @@
...
+      "required": true,
+      "content": {
+        "application/json": {
+          "schema": {
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "properties": {
+              "data": {
```
NickVries, line 27: We should make `collection_id` and `shipment_ids` mandatory properties of data here. It makes no sense to call the endpoint without it.

specification/paths/AddShipmentsToCollection.json
```diff
@@ -0,0 +1,57 @@
...
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "responses": {
+      "204": {
+        "description": "The collection id has been added to the shipments."
```
NickVries, line 53: I'd write this the other way around, since the endpoint is called "add-shipments-to-collection".
```suggestion
        "description": "The shipments have been added to the collection."
```

NickVries (changes requested):

NickVries (approved):

PR #787, opened 2022-12-15
:sparkles: MP-5495 Added item_weight_unit property to shipment items

NickVries:
## Changes
- Introduced `item_weight_unit` for shipment items.

## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-5495
- https://myparcelcombv.atlassian.net/browse/MP-5254

### DIFF
--- a/specification/schemas/BaseShipmentItem.json
+++ b/specification/schemas/BaseShipmentItem.json
@@ -52,6 +52,15 @@
       "example": 135,
       "description": "Weight in grams of a single item within item resource. Should be multiplied by quantity to get total weight."
     },
+    "item_weight_unit": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "example": "mg",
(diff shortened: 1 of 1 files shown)


specification/schemas/BaseShipmentItem.json
```diff
@@ -52,6 +52,15 @@
       "example": 135,
       "description": "Weight in grams of a single item within item resource. Should be multiplied by quantity to get total weight."
     },
+    "item_weight_unit": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "example": "mg",
+      "description": "Unit of measurement for item_weight. Options are: mg, g, kg, oz, lb.",
```
M4tini, line 61: Because of this example, all shipment example jsons show `mg`, but I think the default `g` is a better example.

Also, we can make this attribute an enum, then the description does not have to mention the possible values.

M4tini (changes requested):

PR #790, opened 2023-01-03
:sparkles: Added new rpc endpoint for combined files

zwaans:
# Changes
- deprecated get and get by id endpoints
- removed post endpoint from combined files (can't have duplicate keys in the same file)
- added post RPC endpoint


specification/paths/CombinedFiles.json
```diff
@@ -76,7 +77,7 @@
   },
   "post": {
     "tags": [
-      "CombinedFiles"
+      "RPC"
```
M4tini, line 80: I think we need to keep this (while our app still uses it) and we need to introduce the RPC ass a new endpoint.
Since the RPC format is always `<verb>-<noun>` it can be `POST /combine-files` so it's not a duplicate route.

specification/schemas/CombinedFile.json
```diff
@@ -50,7 +51,6 @@
         "relationships": {
           "type": "object",
           "required": [
-            "shop",
```
M4tini, line 53: Why was this removed?

specification/paths/CombinedFiles.json
```diff
@@ -121,20 +144,12 @@
     },
     "responses": {
       "201": {
```
M4tini, line 146: For the new RPC endpoint, let's make this `200` like the others that return base64 data.

M4tini (changes requested):

M4tini (approved):

specification/paths/CombineFiles.json
```diff
@@ -0,0 +1,83 @@
...
+            }
+          }
+        }
+      }
+    },
+    "responses": {
+      "200": {
+        "description": "Created the combined file.",
+        "content": {
+          "text/csv": {
```
M4tini, line 73: I think this should be `application/pdf`. And I believe `application/zpl` support will be added in the near future, so we could add that as a possible "content" as well, but maybe that's too soon :)

M4tini (changes requested):

specification/schemas/CombinedFile.json
NickVries: Is there a good reason to keep the shop relationship?

M4tini: This schema will not exist anymore once the new RPC endpoint is done, so no need to change anything here.

M4tini (approved):

PR #799, opened 2023-03-03
✨ Added weight_bracket and meta.bracket_price to service-rates

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-5692
---
I decided to expose the 4 database fields as one `weight_bracket` attribute underneath the two existing weight attributes.

### DIFF
--- a/specification/schemas/ServiceRate.json
+++ b/specification/schemas/ServiceRate.json
@@ -15,8 +15,40 @@
             },
             "weight_max": {
               "type": "integer",
-              "example": 2000,
+              "example": 20000,
               "description": "Weight in grams."
+            },
+            "weight_bracket": {
+              "type": "object",
+              "required": [
+                "start",
+                "start_amount",
+                "size",
+                "size_amount"
+              ],
+              "properties": {
+                "start": {
+                  "type": "integer",
+                  "example": 6000,
+                  "description": "Weight in grams."
+                },
+                "start_amount": {
+                  "type": "integer",
+                  "example": 500,
+                  "description": "The starting amount of the bracket price in cents/pence."
+                },
+                "size": {
+                  "type": "integer",
+                  "example": 2000,
+                  "description": "Weight in grams."
+                },
+                "size_amount": {
+                  "type": "integer",
+                  "example": 50,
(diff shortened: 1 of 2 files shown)


specification/schemas/ServiceRate.json
```diff
@@ -15,9 +15,41 @@
...
+                "start_amount": {
+                  "type": "integer",
+                  "example": 500,
+                  "description": "The starting amount of the bracket price in cents/pence."
+                },
+                "size": {
+                  "type": "integer",
+                  "example": 2000,
+                  "description": "Weight in grams."
+                },
```
NickVries, line 44: This should be a bit more informative. What does `size` mean? It's weight in grams, but what does it represent? It took me a while to figure that out, I think our users might even take longer. 

Let's add some information to what it represents, like calling it `bracket_size`, or at least adding info to the description.

specification/schemas/ServiceRate.json
```diff
@@ -15,9 +15,41 @@
...
+                "size_amount": {
+                  "type": "integer",
+                  "example": 50,
+                  "description": "The bracket price will be increased with this amount for as many times as you need to add the bracket `size` to the `start` weight, to cover the weight of the shipment. For this example rate: a shipment of 9 kg would cost 500 + 50 + 50 = 600 cents."
+                }
+              },
+              "description": "Rates with a weight bracket do not have a fixed price between `weight_min` and `weight_max` because their price depends on the actual weight of the parcel. Their price is set to `null` and should be calculated using the bracket attributes (in the same currency as the contract). If you provide `filter[weight]` in the request, you can find the calculated bracket price in the `meta` of this rate."
```
NickVries, line 51:
```suggestion
              "description": "Rates with a weight bracket do not have a fixed price between `weight_min` and `weight_max` because their price depends on the actual weight of the parcel. Their price is set to `null` and should be calculated using the bracket attributes (in the same currency as the contract). If you provide `filter[weight]` in the GET `/service-rates` request, you can find the calculated bracket price in the `meta` of this rate."
```

NickVries (changes requested):

NickVries (approved):

PR #810, opened 2023-05-10
Changed shop relationship to owner relationship for Manifests

zwaans:
https://myparcelcombv.atlassian.net/browse/MP-5319

### DIFF
--- a/specification/paths/Manifests.json
+++ b/specification/paths/Manifests.json
@@ -148,13 +148,14 @@
                         "oneOf": [
                           {
                             "required": [
-                              "shipments"
+                              "shipments",
+                              "owner"
                             ]
                           },
                           {
                             "required": [
                               "shipments",
-                              "owner"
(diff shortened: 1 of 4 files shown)


M4tini (changes requested):
Looks good!

As discussed: let's add `owner` as an include on the GET endpoints and the 3 resources to the `ManifestIncludes` file, so we can easily display it in the portal.

specification/paths/Manifests.json
```diff
@@ -145,9 +145,18 @@
...
+                            "required": [
+                              "shipments"
+                            ]
+                          },
+                          {
+                            "required": [
+                              "shipments",
+                              "owner"
+                            ]
+                          }
```
M4tini, line 159:
```suggestion
                        "required": [
                          "shipments"
```

M4tini (changes requested):

PR #826, opened 2023-06-29
MP-5972 :safety_vest: Introduced return resource

zwaans:


specification/paths/Returns-v1-shops-shop_id-returns.json
```diff
@@ -0,0 +1,195 @@
...
+                      "attributes",
+                      "relationships"
+                    ],
+                    "properties": {
+                      "id": {
+                        "readOnly": true
+                      },
+                      "attributes": {
+                        "required": [
+                          "reference",
```
yoan-myparcel, line 140:
```suggestion
                          "order_reference",
```

yoan-myparcel (changes requested):

PR #830, opened 2023-07-03
:beetle: Removed required id because else the post call requiring an id

zwaans:
# Changes
- Removed required id because else the post call requiring an id

### DIFF
--- a/specification/schemas/ReturnResource.json
+++ b/specification/schemas/ReturnResource.json
@@ -1,8 +1,7 @@
 {
   "type": "object",
   "required": [
-    "type",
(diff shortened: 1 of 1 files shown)


specification/schemas/ReturnResource.json
```diff
@@ -1,8 +1,7 @@
 {
   "type": "object",
   "required": [
-    "type",
-    "id"
+    "type"
```
zwaans, line 4: Removed required id because else the post call requiring an id

yoan-myparcel (approved):

PR #832, opened 2023-07-04
MP-5314 - Added tracking page url to shipment responses

ChrisSercan:
https://myparcelcombv.atlassian.net/browse/MP-5314

### DIFF
--- a/mp/bundle
+++ b/mp/bundle
@@ -11,9 +11,11 @@ echo -e "\033[0;30;47m Injecting vars \033[0m"
 API_HOST=${API_HOST//\//\\/}
 AUTH_HOST=${AUTH_HOST//\//\\/}
 RELEASE_VERSION=${RELEASE_VERSION//\//\\/}
+TRACKING_PAGE_HOST=${TRACKING_PAGE_HOST//\//\\/}
 
 ${COMPOSE} run --rm bundler bash -c "
 sed -i 's/\$API_HOST/${API_HOST}/' dist/swagger.json
 sed -i 's/\$AUTH_HOST/${AUTH_HOST}/' dist/swagger.json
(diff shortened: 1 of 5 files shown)


mp/bundle
```diff
@@ -11,9 +11,11 @@ echo -e "\033[0;30;47m Injecting vars \033[0m"
 API_HOST=${API_HOST//\//\\/}
 AUTH_HOST=${AUTH_HOST//\//\\/}
 RELEASE_VERSION=${RELEASE_VERSION//\//\\/}
+TRACKING_PAGE_HOST=${TRACKING_PAGE_HOST//\//\\/}
 
 ${COMPOSE} run --rm bundler bash -c "
 sed -i 's/\$API_HOST/${API_HOST}/' dist/swagger.json
 sed -i 's/\$AUTH_HOST/${AUTH_HOST}/' dist/swagger.json
 sed -i 's/\$RELEASE_VERSION/${RELEASE_VERSION}/' dist/swagger.json
+sed -i 's/\$TRACKING_PAGE_HOST/${TRACKING_PAGE_HOST}/' dist/swagger.json
```
yoan-myparcel, line 20: This substitution script will work but only locally. We have to also upgrade the Dockerfile that is used to create an image for the live environments deployment -> https://github.com/MyParcelCOM/api-specification/blob/develop/docker/build/entrypoint.sh

In this entrypoint you can see that we do substitution too (although a bit diffirently). Lets pair-program that!

yoan-myparcel (changes requested):

mp/bundle
yoan-myparcel: @ChrisSercan was all clear about the feedback?

yoan-myparcel (approved):

PR #850, opened 2023-07-24
♻️ Change service-options and shipment-statuses to the new resource structure

M4tini:
The new `Resource` and `Response` approach is described here: https://github.com/MyParcelCOM/api-specification/pull/503


yoan-myparcel (approved):
🤷🏻‍♂️ gonna merge it

PR #851, opened 2023-07-24
♻️ Change remaining models to the new resource structure

M4tini:
The new `Resource` and `Response` approach is described here: https://github.com/MyParcelCOM/api-specification/pull/503


specification/paths/Invoices.json
```diff
@@ -149,21 +111,11 @@
...
                           "price"
-                        ],
-                        "properties": {
-                          "price": {
-                            "required": [
-                              "amount",
-                              "currency"
-                            ]
-                          }
-                        }
```
yoerriwalstra, line 163: is price no longer an object with `amount` and `currency` in the new specification, or can you leave out the properties here for a different reason?

M4tini: In this case, we don't have to repeat in this file that `amount` and `currency` are required, because all required properties for a response are now set through the `InvoiceResponse` file.

There are no changes to the final result, stuff is just moved around so there is less repeating going on. As you can see this `InvoiceResponse` is replacing the same bunch of code [here](https://github.com/MyParcelCOM/api-specification/pull/851/files#diff-f5a67c759e4c269e283500f8872b681a299dfbe3bec2e985ea1f26f43d50da21R32) and [here](https://github.com/MyParcelCOM/api-specification/pull/851/files#diff-f5a67c759e4c269e283500f8872b681a299dfbe3bec2e985ea1f26f43d50da21R95) in this PR.

yoerriwalstra (dismissed):

PR #857, opened 2023-07-27
↩️ New return statuses and approve/reject endpoints

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-6006
---
- Replaced `return-concept` status with `return-pending`.
- Added `return-approved` and `return-rejected` statuses.
- Added endpoint to "approve" a return.
- Added endpoint to "reject" a return.


specification/paths/Returns-v1-shops-shop_id-returns-return_id-approve.json
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    "summary": "Approve a return.",
+    "description": "This endpoint modifies the status of a return resource to `return-approved`",
+    "responses": {
+      "204": {
+        "description": "The return is approved."
+      },
+      "404": {
+        "description": "No return was found for the given id."
+      }
```
yoan-myparcel, line 21: We can introduce a 422 in situations when the transition is forbidden because for example the return is already approved (it could be that there are more state transitions that are forbidden)

specification/paths/Returns-v1-shops-shop_id-returns-return_id-reject.json
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    "summary": "Reject a return.",
+    "description": "This endpoint modifies the status of a return resource to `return-rejected`",
+    "responses": {
+      "204": {
+        "description": "The return is rejected."
+      },
+      "404": {
+        "description": "No return was found for the given id."
+      }
```
yoan-myparcel, line 21: Same comment as above about 422

yoan-myparcel (changes requested):

PR #880, opened 2023-08-21
:sparkles: Introduced return method relationship on returns

zwaans:
# Changes
- introduced return method relationship
- Added relationship to returns
- Made relationship optional so it doesn't destroy the tests

### DIFF
--- a/specification/schemas/Return.json
+++ b/specification/schemas/Return.json
@@ -50,6 +50,9 @@
             },
             "shop": {
               "$ref": "#/components/schemas/ShopRelationship"
+            },
+            "return_method": {
+              "$ref": "#/components/schemas/ReturnMethodRelationship"
             }
           }
         }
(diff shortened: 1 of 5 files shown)


specification/schemas/Return.json
```diff
@@ -50,6 +50,9 @@
             },
             "shop": {
               "$ref": "#/components/schemas/ShopRelationship"
+            },
+            "return_method": {
+              "$ref": "#/components/schemas/ReturnMethodRelationship"
             }
```
yoan-myparcel, line 56: We need to make it required!

yoan-myparcel (changes requested):

specification/schemas/Return.json
zwaans: @yoan-myparcel then it will break all return tests until the API is implemented.

zwaans: @yoan-myparcel opened this https://github.com/MyParcelCOM/api-specification/pull/881 so it can be merged after the API gets merged.

yoan-myparcel (approved):

PR #884, opened 2023-09-01
MP-6063 :sparkles: [api-spec] Introduced payment authorization for mollie

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-5795

### DIFF
--- a/specification/tags.json
+++ b/specification/tags.json
@@ -107,6 +107,11 @@
     }
   },
   {
+    "name": "Payments",
+    "externalDocs": {
+    }
+  },
+  {
     "name": "Rate-imports",
     "externalDocs": {
     }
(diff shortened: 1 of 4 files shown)


specification/tags.json
```diff
@@ -106,6 +106,11 @@
     "externalDocs": {
     }
   },
+  {
+    "name": "Payments",
+    "externalDocs": {
+    }
```
yoan-myparcel, line 112: Lets hide this tag by adding it in https://github.com/MyParcelCOM/api-specification/blob/develop/src/myparcelcom.js#L97

yoan-myparcel (changes requested):

yoan-myparcel (approved):

PR #890, opened 2023-09-07
:sparkles: Introduced optional shipment relationship on returns

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-5762

# Changes
- Introduced optional shipment relationship
- Made the other relationships required.

### DIFF
--- a/specification/schemas/Return.json
+++ b/specification/schemas/Return.json
@@ -44,11 +44,6 @@
         "relationships": {
           "type": "object",
           "additionalProperties": false,
-          "required": [
-            "return_method",
-            "shop",
-            "status"
-          ],
           "properties": {
             "return_method": {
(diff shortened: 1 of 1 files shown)


specification/schemas/Return.json
```diff
@@ -44,6 +44,11 @@
         "relationships": {
           "type": "object",
           "additionalProperties": false,
+          "required": [
+            "return_method",
+            "shop",
+            "status"
+          ],
```
M4tini, line 51: Let's define these requirements on the `ReturnResponse` and also on the POST and PATCH endpoint definitions.

The reason for this is more flexibility. If we make `status` required here, then the POST/PATCH body is going to require this relationship as well, while it's actually being set by our API after saving the resource.

M4tini (changes requested):

M4tini (approved):

PR #899, opened 2023-09-15
:wrench: add /returns/{return_id}/request-payment endpoint specification

yoerriwalstra:


specification/paths/Returns-v1-returns-return_id-request-payment.json
```diff
@@ -0,0 +1,62 @@
+{
+  "parameters": [
+    {
+      "$ref": "#/components/parameters/path-return_id"
+    }
+  ],
+  "post": {
+    "tags": [
+      "ReturnPaymentRequests"
```
M4tini, line 9: Since this endpoint starts with `returns/v1/returns/{return_id}` I think we can place it under `Returns`
```suggestion
      "Returns"
```

specification/tags.json
```diff
@@ -138,6 +138,10 @@
     "name": "ReturnMethods",
     "externalDocs": {}
   },
+  {
+    "name": "ReturnPaymentRequests",
+    "externalDocs": {}
+  },
```
M4tini, line 144: That will also remove the need for another header (which we would need to hide)
```suggestion
```

M4tini (changes requested):
You also need to add 2 statuses to the enum, or do you plan to do this in a following PR?
```
return-payment-pending
return-payment-failed
```

specification/paths/Returns-v1-returns-return_id-request-payment.json
yoerriwalstra: done

specification/tags.json
yoerriwalstra: done

yoerriwalstra:
> You also need to add 2 statuses to the enum, or do you plan to do this in a following PR?
> 
> ```
> return-payment-pending
> return-payment-failed
> ```

done

M4tini (approved):

PR #903, opened 2023-09-20
:sparkles: [api-spec] Return Reasons

zwaans:
# Changes
- Introduced return reason resource
- Introduced shops-return reasons to attach a reason to a shop

# Ticket
- https://myparcelcombv.atlassian.net/browse/MP-5756

### DIFF
--- a/specification/paths/Returns-v1-reasons.json
+++ b/specification/paths/Returns-v1-reasons.json
@@ -130,7 +130,6 @@
                       "attributes": {
                         "required": [
                           "description",
-                          "tags",
                           "accepts_attachments",
                           "requires_attachments"
                         ]
--- a/specification/paths/Returns-v1-reasons.json
+++ b/specification/paths/Returns-v1-reasons.json
@@ -1,0 +1,174 @@
+{
+  "get": {
+    "tags": [
+      "ReturnReasons"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
+        ]
+      }
+    ],
+    "summary": "Get all available return reasons.",
+    "description": "This endpoint retrieves returns reasons available to your API client.",
+    "parameters": [
+      {
+        "name": "include",
+        "in": "query",
+        "description": "Comma separated string of the relationship names you want to include the data of.<br>The relationships that can be included are:<ul><li>`shop`</li></ul>",
+        "schema": {
+          "type": "string"
+        }
+      },
+      {
+        "$ref": "#/components/parameters/query-filter-tags"
+      }
+    ],
+    "responses": {
+      "200": {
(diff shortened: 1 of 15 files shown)


specification/schemas/ReturnReason.json
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "relationships": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "shop": {
+              "$ref": "#/components/schemas/ShopRelationship"
+            }
+          }
+        }
```
M4tini, line 41: ReturnReasons are generic things that our customers (shops) can decide to activate/use. One ReturnReason can be used by 0 or more shops. Only the linked ReturnReasons will be returned by the shop-specific endpoint.

If we would expose any relationship on a ReturnReason, I think it should be an `owner` relationship with us (the broker) as owner. But I don't think this is needed yet.

specification/paths/Returns-v1-shops-shop_id-reasons.json
```diff
@@ -0,0 +1,122 @@
...
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/ReturnReasonResponse"
+                    },
+                    {
+                      "required": [
+                        "relationships"
+                      ]
+                    }
+                  ]
```
M4tini, line 110:
```suggestion
                  "$ref": "#/components/schemas/ReturnReasonResponse"
```

M4tini (changes requested):

specification/paths/Returns-v1-reasons.json
```diff
@@ -0,0 +1,174 @@
+{
+  "get": {
+    "tags": [
+      "ReturnReasons"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
```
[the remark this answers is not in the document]
zwaans: I don't think the `return_reasons.manage` scope is needed here. If someone wants to attach the reasons to his shop he needs to be able to view all the return reasons to be able to select the ones that he wants to attach to the shop right?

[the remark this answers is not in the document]
M4tini: I thought if someone only has `return_reasons.manage` and not `returns.manage` then he should be able to GET the resources they have created. I guess this will never happen and our users will have both scopes.

Anyway, it sounds logical that `return_reasons.manage` should give access to all CRUD endpoints of ReturnMethods.

specification/schemas/ReturnReason.json
M4tini: I think we should 🔥 this shop relationship

M4tini (approved):

PR #910, opened 2023-09-25
:beetle: Remove required return reason relationships on post

zwaans:
# Changes
- No relationships are required when creating a return reason


M4tini (changes requested):
This file has more shop relationship stuff that needs to be 🔥

M4tini (approved):

PR #911, opened 2023-09-25
:lipstick: Fix POST shop reasons endpoint

zwaans:
# Changes
- Added the correct request body for the post shop reasons endpoint

### DIFF
--- a/specification/paths/Returns-v1-shops-shop_id-reasons.json
+++ b/specification/paths/Returns-v1-shops-shop_id-reasons.json
@@ -53,10 +53,10 @@
         ]
       }
     ],
-    "summary": "Link a return reason to the given shop.",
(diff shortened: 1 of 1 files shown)


specification/paths/Returns-v1-shops-shop_id-reasons.json
```diff
@@ -53,10 +53,10 @@
         ]
       }
     ],
-    "summary": "Link a return reason to the given shop.",
-    "description": "This endpoint links a return reason to the specified shop.",
+    "summary": "Link return reason's to the given shop.",
```
M4tini, line 56:
```suggestion
    "summary": "Link return reasons to the given shop.",
```

M4tini (changes requested):

M4tini (approved):

PR #924, opened 2023-10-19
♻️ ReturnMethod default_weight is not always required

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-6141
---
- Remove the requirement of `default_weight`.
- Added description to indicate when it is required.

### DIFF
--- a/specification/paths/Returns-v1-return-methods.json
+++ b/specification/paths/Returns-v1-return-methods.json
@@ -129,9 +129,13 @@
                       },
                       "attributes": {
                         "required": [
-                          "return_type",
-                          "default_weight"
-                        ]
+                          "return_type"
+                        ],
+                        "properties": {
+                          "default_weight": {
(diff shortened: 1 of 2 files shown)


yoan-myparcel:
@M4tini shall we draft this because staging is not merged yet?

specification/paths/Returns-v1-return-methods.json
```diff
@@ -129,9 +129,13 @@
...
-                        ]
+                          "return_type"
+                        ],
+                        "properties": {
+                          "default_weight": {
+                            "description": "Required if the `return_type` is set to `shipment`."
+                          }
+                        }
                       },
                       "relationships": {
```
yoerriwalstra, line 140: the relationships `contract` and `service` are also required if the `return_type = shipment`. could you add a similar comment to those properties to indicate this to the user?

yoerriwalstra (approved):

PR #933, opened 2023-10-31
:lipstick: Removed 'office' as a prefix from the get pudo by id endpoint

zwaans:


M4tini (changes requested):
I thought we would remove it from all PUDO endpoint titles?

M4tini (approved):

PR #943, opened 2023-11-10
:sparkles: add new hook action types

yoerriwalstra:
**Related Jira issue**
https://myparcelcombv.atlassian.net/browse/MP-6213

### DIFF
--- a/specification/schemas/HookActionApproveReturn.json
+++ b/specification/schemas/HookActionApproveReturn.json
@@ -10,21 +10,6 @@
       "type": "string",
       "pattern": "^approve-return$",
       "example": "approve-return"
-    },
-    "values": {
-      "type": "array",
-      "items": {
-        "type": "object",
-        "required": [
-          "return_id"
-        ],
-        "additionalProperties": false,
-        "properties": {
-          "return_id": {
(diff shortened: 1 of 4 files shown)


specification/schemas/HookActionApproveReturn.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "string",
+      "pattern": "^approve-return$",
+      "example": "approve-return"
+    },
+    "values": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "return_id"
```
yoan-myparcel, line 19: I do not understand why would return_id be an accepted value here. We are applying these actions on any return resource that matches the predicates.

yoan-myparcel (changes requested):

specification/schemas/HookActionApproveReturn.json
M4tini: Yep, the entire `values` part can be dropped. It will approve/reject the (return) resource that triggered the hook.

yoerriwalstra: fixed

yoan-myparcel (approved):

PR #954, opened 2023-11-22
:sparkles: Added new hook action for requesting payments + checkout URL jump route

yoan-myparcel:
# Changes
Alongside the new hook action, I also added the endpoint that will generate the mollie checkout URL once visited.

### DIFF
--- a/specification/paths/Returns-v1-returns-return_id-payment-checkout-url-payment_id.json
+++ b/specification/paths/Returns-v1-returns-return_id-payment-checkout-url-payment_id.json
@@ -39,7 +39,7 @@
         "description": "The payment provider does not allow the return payment link to be created or HMAC signature is invalid."
       },
       "422": {
-        "description": "The payment link expired."
+        "description": "When not connected to the payment provider, or payment is already paid"
       },
       "404": {
         "description": "The payment was not found."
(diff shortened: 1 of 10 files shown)


specification/paths/Returns-v1-returns-return_id-payment-checkout-url-payment_id.json
```diff
@@ -0,0 +1,52 @@
...
+      },
+      "403": {
+        "description": "The payment provider does not allow the return payment link to be created or HMAC signature is invalid."
+      },
+      "422": {
+        "description": "The payment link expired."
+      },
+      "404": {
+        "description": "The payment was not found."
+      },
```
yoerriwalstra, line 46: Just looking at the Controller action `ReturnPaymentController::checkoutUrl` (so not tracking down possible exceptions in different files), we currently throw a 422 exception when:
1. the Shop is not connected to Mollie
2. the ReturnPayment is not found
3. the ReturnPayment is already paid

If the payment link has expired, we just create a new one (ReturnPaymentController line 92-96). 

I think it makes more sense to throw a ModelNotFoundException when the ReturnPayment is not found in the API, so this part of the spec would stay valid (after the change in the API)

yoan-myparcel: That is correct! I will change the description

yoan-myparcel: @yoerriwalstra specified the description for case 1 and 3. 2 is 404

yoerriwalstra (approved):

PR #971, opened 2023-12-31
🚚 Organize schemas

M4tini:
🪄 Experimental Days
---
- 🚚 Organized schemas into subfolders to avoid scrolling through a massive directory.
- 🔥 Removed a few unused or unnecessary schemas.


yoan-myparcel:
@M4tini Can you confirm that the unused schemas you removed are:
- Day
- FinalMileCarrier
- Language

M4tini:
@yoan-myparcel Yes:
- `Day` was only used by the Service schema, so I moved the contents there.
- `FinalMileCarrier` was only used by the Shipment schema, so I moved the contents there.
- `Language` was not used at all (used to be on the organization or user, but we removed that in favor of `Locale`).

yoan-myparcel (approved):

PR #973, opened 2024-01-05
[api-spec] :sparkles: Introduced return item analytics

zwaans:


specification/schemas/returns/MostReturnedItem.json
```diff
@@ -0,0 +1,53 @@
...
+        },
+        "amount": {
+          "type": "number"
+        },
+        "return_reasons": {
+          "type": "array",
+          "required": [
+            "code",
+            "name",
+            "amount"
```
yoan-myparcel, line 30: Somehow these are not taken into account. Perhaps then need to go under "items"?
image

yoan-myparcel (changes requested):

yoan-myparcel (approved):

PR #974, opened 2024-01-08
✨  Payment providers

yoan-myparcel:
# What's new
- `/payments/v1/providers?shop_id={shop_id}`
- `/payments/v1/providers/{code}/connect`
- `/payments/v1/providers/{code}/disconnect`

### DIFF
--- a/specification/paths/Payments-v1-providers-code-connect.json
+++ b/specification/paths/Payments-v1-providers-code-connect.json
@@ -1,7 +1,7 @@
 {
   "post": {
     "tags": [
-      "Payments"
+      "PaymentProviders"
     ],
     "security": [
       {
@@ -10,10 +10,10 @@
         ]
       }
     ],
-    "summary": "Request an authorization url from Mollie.",
-    "description": "This endpoint returns an authorization url to connect Mollie to your shop.",
+    "summary": "Connect to a payment provider.",
+    "description": "This endpoint connects a payment provider to your shop.",
     "requestBody": {
-      "description": "The shop ID to connect with.",
+      "description": "The shop ID and other settings to use for the connection.",
       "required": true,
       "content": {
         "application/vnd.api+json": {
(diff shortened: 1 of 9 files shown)


specification/paths/Payments-v1-providers-code-connect.json
```diff
@@ -36,6 +36,16 @@
                     "format": "uuid",
                     "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
                     "example": "4067e35b-09cc-42b9-8553-6eaf1788df35"
+                  },
+                  "settings": {
+                    "type": "object",
+                    "description": "The settings to use for the connection. See the payment provider's settings_format for details."
+                  },
```
M4tini, line 43: We could add an `"example"` here, so the preview panel does not just show `"settings": { },`. We do the same for our shop-integration routes. I'm also fine leaving it like this, since it's an optional attribute.
What do yo think?

specification/paths/Payments-v1-providers.json
```diff
@@ -0,0 +1,84 @@
...
+    "summary": "Get all payment providers.",
+    "description": "This endpoint retrieves payment providers available to your API client.",
+    "parameters": [
+      {
+        "name": "shop_id",
+        "in": "query",
+        "description": "When specified only payment providers connected to the given shop are returned.",
+        "schema": {
+          "type": "string"
+        }
```
M4tini, line 22: Let's use the existing shop filter, which we also use on our `/returns/v1/returns` endpoint.

```suggestion
        "$ref": "#/components/parameters/query-filter-shop"
```

specification/schemas/payment-providers/PaymentProviderResource.json
```diff
@@ -0,0 +1,19 @@
...
+    "type": {
+      "type": "string",
+      "pattern": "^payment-providers$",
+      "example": "payment-providers"
+    },
+    "id": {
+      "type": "string",
+      "format": "uuid",
+      "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
+      "example": "7fac478b-b3c4-46e6-b29e-a31599d27a7a"
```
M4tini, line 16: This UUID is already used for our Return resource. Let's generate a fresh one for this resource.

```suggestion
      "example": "6718621d-521c-4045-b4ef-1ab4dc29b316"
```

M4tini (changes requested):

specification/paths/Payments-v1-providers-code-connect.json
yoan-myparcel: Example will be very useful! I will add one!

specification/paths/Payments-v1-providers.json
yoan-myparcel: It is not really a filter though, I prefer we stick with `shop_id` query parameter because the intended behavior is different that the filter

M4tini (approved):

PR #981, opened 2024-01-15
👌🏻 Payment providers - QA feedback

yoan-myparcel:
# What's new?
- Add additional error responses for payment provider connect endpoint
- Changed behavior of shop_id when getting providers

### DIFF
--- a/specification/schemas/payment-providers/PaymentProviderResponse.json
+++ b/specification/schemas/payment-providers/PaymentProviderResponse.json
@@ -19,7 +19,14 @@
             "logo_url",
             "connects_through_oauth2",
             "created_at"
-          ]
+          ],
+          "properties": {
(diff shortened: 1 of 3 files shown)


specification/schemas/payment-providers/PaymentProviderResponse.json
```diff
@@ -19,7 +19,14 @@
             "logo_url",
             "connects_through_oauth2",
             "created_at"
-          ]
+          ],
+          "properties": {
+            "connected": {
```
NickVries, line 24: Let's put this property in the `meta`, since it's not information about the payment provider itself, but rather side information based on request parameters.

NickVries (changes requested):

NickVries (approved):

PR #990, opened 2024-01-26
⚡  Added shipment metadata for ReturnSimpleResponse

yoan-myparcel:
# What's new?
When getting multiple returns, a meta field is added for each which contains the shipment (if any) tracking code and status code.

Removed the shipment include when getting multiple returns.

### DIFF
--- a/specification/schemas/returns/ReturnSimpleResponse.json
+++ b/specification/schemas/returns/ReturnSimpleResponse.json
@@ -57,7 +57,6 @@
               "type": "object",
               "additionalProperties": false,
               "required": [
-                "tracking_code",
                 "status"
               ],
               "properties": {
--- a/specification/schemas/returns/ReturnSimpleResponse.json
+++ b/specification/schemas/returns/ReturnSimpleResponse.json
@@ -54,9 +54,7 @@
           "additionalProperties": false,
           "properties": {
             "shipment_tracking_code": {
-              "type": "string",
-              "description": "Tracking code of the shipment",
-              "example": "3SABCD0123456789"
+              "$ref": "#/components/schemas/TrackingCode"
             },
             "shipment_status_code": {
               "type": "string",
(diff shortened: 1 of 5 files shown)


specification/schemas/returns/ReturnSimpleResponse.json
```diff
@@ -47,6 +47,56 @@
...
+        "meta": {
+          "readOnly": true,
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "shipment": {
+              "type": "object",
+              "additionalProperties": false,
+              "required": [
+                "tracking_code",
```
M4tini, line 60: If the shipment is registration failed, there is a status but no tracking code
```suggestion
```

specification/schemas/returns/ReturnSimpleResponse.json
```diff
@@ -47,6 +47,56 @@
...
+                  "description": "Tracking code of the shipment",
+                  "example": "3SDFGDFGDFGDFG"
+                },
+                "status": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "code": {
+                      "type": "string",
+                      "enum": [
```
M4tini, line 75: Instead of defining the entire enum again, let's make a `StatusCode` schema and use that in both places:
```
"code": {
  "$ref": "#/components/schemas/StatusCode"
}
```

M4tini (changes requested):

specification/schemas/returns/ReturnSimpleResponse.json
M4tini: Should we also expose the `tracking_url`? I assume when showing a tracking code somewhere that we want to make it link to the tracking website.

specification/schemas/returns/ReturnSimpleResponse.json
yoan-myparcel: That wouldn't be correct because the StatusCode schema includes status codes for other resources, other than shipments.

The resulting duplication is very small and imo manageable.

M4tini (approved):

M4tini (approved):

PR #997, opened 2024-02-07
👌🏻  Added missing image_url for most returned item analytics

yoan-myparcel:


specification/schemas/returns/MostReturnedItem.json
```diff
@@ -12,6 +12,7 @@
       "required": [
         "sku",
         "name",
+        "image_url",
```
M4tini, line 15: This cannot be required, since it's optional for our Return resource items.
```suggestion
```

M4tini (changes requested):

specification/schemas/returns/MostReturnedItem.json
M4tini: I see that the API is going to [default to a confetti SVG](https://github.com/MyParcelCOM/api/pull/3182/files#diff-36283a1efa4f84c3a5afdbab4f0006cbca5d8af147318fee511dbf1bd042b812R19) but placeholders are a responsibility of the front-end.

M4tini (approved):

NickVries (approved):

PR #1007, opened 2024-03-04
:sparkles: update ReturnReason schemas to include possible description object

yoerriwalstra:
**Related issue**
https://myparcelcombv.atlassian.net/browse/MP-6471

### DIFF
--- a/specification/paths/Returns-v1-reasons.json
+++ b/specification/paths/Returns-v1-reasons.json
@@ -20,6 +20,15 @@
     "parameters": [
       {
         "$ref": "#/components/parameters/query-filter-tags"
+      },
+      {
+        "in": "header",
+        "name": "Accept-Language",
+        "schema": {
+          "type": "string"
+        },
+        "description": "Description of your optional header",
+        "required": false
       }
     ],
     "responses": {
@@ -110,7 +119,7 @@
               "data": {
                 "allOf": [
                   {
-                    "$ref": "#/components/schemas/ReturnReason"
+                    "$ref": "#/components/schemas/CreateReturnReason"
                   },
                   {
                     "required": [
@@ -122,6 +131,7 @@
                       },
                       "attributes": {
                         "required": [
+                          "code",
                           "description",
                           "accepts_attachments",
                           "requires_attachments",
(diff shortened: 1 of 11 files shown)


specification/paths/Returns-v1-reasons.json
```diff
@@ -20,6 +20,15 @@
...
         "$ref": "#/components/parameters/query-filter-tags"
+      },
+      {
+        "in": "header",
+        "name": "Accept-Language",
+        "schema": {
+          "type": "string"
+        },
+        "description": "Description of your optional header",
+        "required": false
```
M4tini, line 31:
```suggestion
        "description": "Determines the language in which the ``description`` is returned,"
```
and maybe something about "ISO 3166-1 alpha-2 compliant language tag" like you mentioned below

specification/schemas/return-reasons/ReturnReason.json
```diff
@@ -15,8 +15,17 @@
...
-              "example": "The item isn't what I expected it to be."
+              "oneOf": [
+                {
+                  "$ref": "#/components/schemas/ReturnReasonManagementDescription"
+                },
+                {
+                  "type": "string",
+                  "description": "A localized return reason description",
+                  "example": "The item isn't what I expected it to be."
+                }
```
M4tini, line 26: I would swap these `oneOf` elements, so the JSON example shows the "string" version by default. This is more clear for regular people using the endpoint (the management version is mainly for us).

M4tini (changes requested):
I think the following endpoints should implement the header and default to `en-GB` as well.
- `/returns/v1/reasons/{return_reason_id}`
- `/returns/v1/shops/{shop_id}/reasons`

specification/paths/Returns-v1-reasons.json
yoerriwalstra: done

specification/schemas/return-reasons/ReturnReason.json
yoerriwalstra: done

M4tini (changes requested):
Let's also add the header to `/returns/v1/reasons/{return_reason_id}` and move the header to a reusable `#/components/parameters/header-accept-language`

yoerriwalstra:
> Let's also add the header to /returns/v1/reasons/{return_reason_id} and move the header to a reusable #/components/parameters/header-accept-language

Done

M4tini (approved):

PR #1014, opened 2024-03-08
✨  Filters: shipping and return/litb return methods

yoan-myparcel:
# Changes
Decided not to add separate parameter files for these two filters because they are not reused anywhere else.


specification/paths/Returns-v1-return-methods.json
```diff
@@ -26,6 +26,16 @@
...
+      },
+      {
+        "name": "filter[shipping_and_return]",
+        "in": "query",
+        "description": "When set, the endpoint will return only shipping return methods that offer shipping and return service."
+      },
+      {
+        "name": "filter[label_in_the_box]",
+        "in": "query",
+        "description": "When set, the endpoint will return only shipping return methods that offer label in the box service."
```
yoan-myparcel, line 38: Yes, intentionally did not set a schema. For me it is enough if the `filter[...]` is set in the query, no value needed.

specification/paths/Returns-v1-return-methods.json
```diff
@@ -26,6 +26,22 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
+        "name": "filter[shipping_and_return]",
+        "in": "query",
+        "description": "When set to true, the endpoint will return only shipping return methods that offer shipping and return service. However, in case that there is no shipping return method available, the endpoint will respond with all return methods, except label in the box, unless specified as a filter. When set to false, the endpoint will exclude all shipping and return return methods.",
```
NickVries, line 33: Couple of things:
1. If I were a user of our product, I'd want to know what identifies a shipping and return (or label in the box) service. I mean technically, is there a flag on the service that identifies it, or is it a specific service code etc.
2. If there are no return methods with shipping and return service, shouldn't we simply return an empty array?
3. If there are no shiping and return methods, why are we returning all _but_ the label in the box return methods? What specifically links these two together?

NickVries (changes requested):

NickVries (approved):

PR #1017, opened 2024-03-21
:sparkles: add `return_portal_locales` attribute to Shop

yoerriwalstra:
**Related issue**
https://myparcelcombv.atlassian.net/browse/MP-6506

### DIFF
--- a/specification/schemas/shops/Shop.json
+++ b/specification/schemas/shops/Shop.json
@@ -140,6 +140,14 @@
                   "description": "The default language that is used for emails and webpages for this shop."
                 }
               ]
+            },
+            "return_portal_locales": {
+              "type": "array",
+              "items": {
+                "$ref": "#/components/schemas/Locale"
+              },
+              "example": "nl-NL",
+              "description": "The languages that are available to your consumer when they visit the Return portal. Defaults to your shop's locale. The selected locale is used for emails related to returns."
             }
           }
         },


specification/schemas/locales/Locale.json
```diff
@@ -0,0 +1,26 @@
...
+            "code": {
+              "$ref": "#/components/schemas/LocaleEnum"
+            },
+            "name": {
+              "type": "string",
+              "example": "English"
+            },
+            "created_at": {
+              "$ref": "#/components/schemas/Timestamp"
+            }
```
yoan-myparcel, line 20: I'd remove that since I cannot think of a reason to have it exposed!

specification/schemas/shops/Shop.json
```diff
@@ -132,14 +132,22 @@
...
                   "description": "The default language that is used for emails and webpages for this shop."
                 }
               ]
+            },
+            "return_portal_locales": {
+              "type": "array",
+              "items": {
+                "$ref": "#/components/schemas/LocaleEnum"
+              },
+              "example": "nl-NL",
```
yoan-myparcel, line 149: Example appears to be wrong, should be an array
```suggestion
              "example": ["nl-NL"],
```

yoan-myparcel (changes requested):

specification/schemas/locales/Locale.json
yoerriwalstra: removed

specification/schemas/shops/Shop.json
yoerriwalstra: fixed

yoan-myparcel (approved):

PR #1018, opened 2024-03-22
:sparkles: Introduced invitations resource

zwaans:
# Related ticket
- https://myparcelcombv.atlassian.net/browse/MP-6448

### DIFF
--- a/specification/schemas/invitations/InvitationResponse.json
+++ b/specification/schemas/invitations/InvitationResponse.json
@@ -1,0 +1,59 @@
+{
+  "allOf": [
+    {
+      "$ref": "#/components/schemas/Invitation"
+    },
+    {
+      "required": [
+        "id",
+        "attributes",
+        "relationships",
+        "links"
+      ],
+      "properties": {
+        "attributes": {
+          "required": [
+            "email"
+          ],
+          "expires_at": {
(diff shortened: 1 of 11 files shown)


specification/schemas/invitations/InvitationResponse.json
```diff
@@ -0,0 +1,59 @@
...
+              "example": "Sherlock"
+            },
+            "last_name": {
+              "type": "string",
+              "example": "Holmes"
+            },
+            "email": {
+              "$ref": "#/components/schemas/Email",
+              "example": "s.holmes@holmesinvestigations.com"
+            }
```
NickVries, line 33: These properties are already defined in the Invitation schema, no need to redefine them.

specification/schemas/invitations/InvitationResponse.json
```diff
@@ -0,0 +1,59 @@
...
+        "links"
+      ],
+      "properties": {
+        "attributes": {
+          "required": [
+            "email"
+          ],
+          "expires_at": {
+            "readOnly": true
+          },
```
NickVries, line 20: I think this is improperly indented. Should be under `properties`.

NickVries (changes requested):

PR #1032, opened 2024-04-26
:ghost: add hide_customer_reference and hide_sender_address attributes

yoerriwalstra:
## Related issue
https://myparcelcombv.atlassian.net/browse/MP-6559

### DIFF
--- a/specification/schemas/shops/Shop.json
+++ b/specification/schemas/shops/Shop.json
@@ -107,15 +107,27 @@
                 "tracking": {
                   "allOf": [
                     {
+                      "$ref": "#/components/schemas/Style"
+                    },
+                    {
                       "properties": {
                         "logo_url": {
                           "$ref": "#/components/schemas/ImageUrl"
+                        },
+                        "hide_customer_reference": {
+                          "type": "boolean",
+                          "example": false,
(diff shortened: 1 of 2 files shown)


specification/schemas/shops/Shop.json
```diff
@@ -106,16 +106,28 @@
...
+                          "type": "boolean",
+                          "example": false,
+                          "default": false,
+                          "description": "Whether to hide the customer reference from the shipment's tracking page."
+                        },
+                        "hide_sender_address": {
+                          "type": "boolean",
+                          "example": false,
+                          "default": false,
+                          "description": "Whether to hide the sender address from the shipment's tracking page."
```
NickVries, line 127: It seems that these properties will always be in the response. Can you make the spec reflect that by putting these in the `required` array?

NickVries (changes requested):

specification/schemas/shops/Shop.json
yoerriwalstra: made boolean style properties (for both `returns` and `tracking`) `required` in the ShopResponse

NickVries (approved):

PR #1038, opened 2024-05-08
:wrench: [api-spec] outbound shipment delivered at

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-6508

### DIFF
--- a/specification/schemas/returns/ReturnResponse.json
+++ b/specification/schemas/returns/ReturnResponse.json
@@ -184,6 +184,18 @@
               "example": "$API_HOST/returns/v1/returns/8ddfb83f-d600-49c7-b10e-66dd4dc7686a"
             }
           }
+        },
+        "meta": {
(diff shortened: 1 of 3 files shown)


yoerriwalstra:
shall we also add the new hook trigger operators - `older_than`, `newer_than` - to this PR (in `HookTrigger.json`), or make a separate one for that?

yoan-myparcel (approved):

yoan-myparcel:
@yoerriwalstra to unblock the API PR I suggest we do a new PR only for the hook trigger operators

yoerriwalstra:
> @yoerriwalstra to unblock the API PR I suggest we do a new PR only for the hook trigger operators

sure, the new operators are already mentioned in the last [API subtask](https://myparcelcombv.atlassian.net/browse/MP-6583), so makes sense to couple the api-spec with that ticket

PR #1039, opened 2024-05-13
MP-6590 🔍  Added search.name filter to shop and organization endpoints.

NickVries:
## Changes
- ✨ Introduced `filter[search.name]` on shops and organizations get endpoints.

## Related issues
- https://myparcelcombv.atlassian.net/browse/MP-6590
- https://myparcelcombv.atlassian.net/browse/MP-6086

### DIFF
--- a/specification/paths/Organizations.json
+++ b/specification/paths/Organizations.json
@@ -31,7 +31,7 @@
       {
         "name": "filter[search.name]",
         "in": "query",
-        "description": "String of characters you want to look for in a organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
+        "description": "String of characters you want to look for in an organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
         "schema": {
           "type": "string"
         }
--- a/specification/paths/Organizations.json
+++ b/specification/paths/Organizations.json
@@ -31,7 +31,7 @@
       {
         "name": "filter[search.name]",
         "in": "query",
-        "description": "String of characters you want to look for in a shop's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
+        "description": "String of characters you want to look for in a organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
         "schema": {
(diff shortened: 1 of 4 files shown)


specification/paths/Organizations.json
```diff
@@ -27,6 +27,14 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
+        "name": "filter[search.name]",
+        "in": "query",
+        "description": "String of characters you want to look for in a shop's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
```
NickVries, line 34:
```suggestion
        "description": "String of characters you want to look for in a organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
```

specification/paths/Organizations.json
```diff
@@ -27,6 +27,14 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
+        "name": "filter[search.name]",
+        "in": "query",
+        "description": "String of characters you want to look for in a organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
```
NickVries, line 34:
```suggestion
        "description": "String of characters you want to look for in an organization's name. Similar to the `filter[search]` parameter, but searches only in the `name` attribute.",
```

specification/paths/Organizations.json
```diff
@@ -27,6 +27,14 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
```
zwaans, line 31: Can we place this in its own file so we can reference it instead of declaring it twice? This will introduce some issues with the description I guess? What do you think.

zwaans (changes requested):

specification/paths/Organizations.json
NickVries: I was thinking the same thing, also with the issues with the description.

But I just thought that perhaps instead of saying that we search in the shop/organization's name, we can just say that we search in the resource's name attribute. wdyt?

zwaans (approved):

PR #1045, opened 2024-06-04
:sparkles: introduced return outcomes on shop, removed docker version

zwaans:


specification/schemas/shops/Shop.json
```diff
@@ -161,8 +161,27 @@
...
+                "refund": {
+                  "type": "boolean",
+                  "default": "true",
+                  "example": "true",
+                  "readOnly": true,
+                  "description": "The shop offers refund as a return outcome, this property cannot be changed."
+                },
+                "exchange": {
+                  "type": "boolean",
+                  "example": "true",
```
M4tini, line 181: Not a lot of them will have this + good to not have all boolean examples show true
```suggestion
                  "example": "false",
```

M4tini (changes requested):

M4tini (approved):

PR #1053, opened 2024-07-01
✨ Introduce broker filters on endpoints for shipments and reports

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-3738
---
To be able to filter the shipment overview and the reports on the broker(s) that are selected (for multi-broker admins).

Meanwhile we also need to be able to limit the organization + shop filters to these broker for the values to make sense.

**Question**
Currently the carriers / contracts / services endpoints are internally filtered on all brokers the user + client have access to.
I guess we should also allow a `filter[broker]` to be passed via the URL so these values can be limited to just 1 broker?

### DIFF
--- a/specification/paths/Organizations.json
+++ b/specification/paths/Organizations.json
@@ -30,6 +30,9 @@
       },
       {
         "$ref": "#/components/parameters/query-filter-search-name"
+      },
+      {
+        "$ref": "#/components/parameters/query-filter-broker"
       }
     ],
     "responses": {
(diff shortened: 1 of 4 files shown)


yoan-myparcel:
@M4tini shall we update search-shipments too?

M4tini:
Not sure where to update `search-shipments` and I'd say we leave it since the implementation is still not concrete

yoan-myparcel:
@M4tini fair point, lets undraft this then

yoan-myparcel (approved):

PR #1057, opened 2024-07-12
🚑️ Post brokers without specifying a domain

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-6734


specification/schemas/brokers/Broker.json
```diff
@@ -15,9 +15,12 @@
...
             },
             "domain": {
-              "type": "string",
+              "type": [
+                "string",
+                "null"
+              ],
               "maxLength": 255,
-              "example": "app.sendmyparcel.com"
+              "example": "app.myparcel.com"
```
yoan-myparcel, line 23: Lets have a different example, like `example.com`, as in to motivate users that here they can enter their custom domain name. I think a description will be of use too.

```suggestion
              "example": "example.com",
              "description": "Set a custom domain for the backoffice. Defaults to app.myparcel.com."
```

yoan-myparcel (changes requested):

yoan-myparcel (approved):

PR #1059, opened 2024-07-16
:sparkles: Manage demo orders specification

yoerriwalstra:
## Changes
- add specification for GET and POST `/returns/v1/demo-orders`
- add specification for GET `/returns/v1/demo-orders/{demo_order_id}`

## Related issue
https://myparcelcombv.atlassian.net/browse/MP-6732

### DIFF
--- a/specification/paths/Returns-v1-demo-orders.json
+++ b/specification/paths/Returns-v1-demo-orders.json
@@ -34,7 +34,6 @@
               "type": "object",
               "required": [
                 "data",
-                "meta",
                 "links"
               ],
               "properties": {
--- a/specification/paths/Returns-v1-demo-orders.json
+++ b/specification/paths/Returns-v1-demo-orders.json
@@ -1,0 +1,195 @@
+{
+  "get": {
+    "tags": [
+      "Returns"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
+        ]
+      }
+    ],
+    "summary": "Get all demo orders resources.",
+    "description": "This endpoint retrieves generated orders to test creating returns.",
+    "parameters": [
+      {
+        "$ref": "#/components/parameters/query-filter-shop"
+      },
+      {
+        "name": "include",
+        "in": "query",
+        "description": "Comma separated string of the relationship names you want to include the data of.<br> The relationships that can be included are:<ul><li>`shop`</li></ul>",
+        "schema": {
+          "type": "string"
+        }
+      }
+    ],
+    "responses": {
+      "200": {
+        "description": "Retrieved the demo orders.",
+        "content": {
+          "application/vnd.api+json": {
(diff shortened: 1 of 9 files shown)


specification/schemas/return-orders/DemoOrderResponse.json
```diff
@@ -0,0 +1,73 @@
...
+              "type": "boolean"
+            },
+            "label_in_the_box": {
+              "type": "boolean"
+            },
+            "package_type": {
+              "$ref": "#/components/schemas/PackageType"
+            }
+          }
+        }
```
yoan-myparcel, line 69: These meta properties are not necessary, they actually come from the carrier integration using the outbound shipment identifier. 
```suggestion
        }
```

yoan-myparcel (changes requested):

specification/paths/Returns-v1-demo-orders.json
```diff
@@ -0,0 +1,194 @@
...
+            "properties": {
+              "data": {
+                "allOf": [
+                  {
+                    "$ref": "#/components/schemas/DemoOrder"
+                  },
+                  {
+                    "required": [
+                      "attributes",
+                      "relationships"
```
yoan-myparcel, line 123: `relationships` is set as required, but not defined

specification/schemas/return-orders/DemoOrder.json
```diff
@@ -0,0 +1,45 @@
...
+    {
+      "$ref": "#/components/schemas/DemoOrderResource"
+    },
+    {
+      "properties": {
+        "type": {
+          "type": "string",
+          "pattern": "^demo-orders$",
+          "example": "demo-orders"
+        },
```
yoan-myparcel, line 12: No longer necessary, it's defined in the DemoOrderResource
```suggestion
```

specification/schemas/return-orders/DemoOrderResponse.json
```diff
@@ -0,0 +1,51 @@
+{
+  "allOf": [
+    {
+      "$ref": "#/components/schemas/DemoOrder"
+    },
+    {
+      "required": [
+        "id",
+        "attributes",
+        "relationships"
```
yoan-myparcel, line 10: same here, relationships is set as required but definition is missing.

yoan-myparcel (changes requested):

specification/schemas/return-orders/DemoOrderResponse.json
yoerriwalstra: I believe adding the `shop` relationship on the DemoOrder solves both cases of missing relationships, right? Since DemoOrderResponse uses `allOf: [ { "$ref": ".../schemaPath/DemoOrder.json"  } ]`

yoan-myparcel (approved):

PR #1073, opened 2024-08-02
✨ Add volumetric_weight_divisor_factor to Contract resource

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-6757
---
- Added `volumetric_weight_divisor_factor` float to the `Contract` resource.
- Adjusted existing number types to be defined according to [the OpenAPI Number definition](https://swagger.io/docs/specification/data-models/data-types/#numbers).
  > OpenAPI has two numeric types, `number` and `integer`, where number includes both integer and floating-point numbers. An optional `format` keyword serves as a hint for the tools to use a specific numeric type.

### DIFF
--- a/specification/schemas/contracts/Contract.json
+++ b/specification/schemas/contracts/Contract.json
@@ -20,8 +20,9 @@
               "$ref": "#/components/schemas/Currency"
             },
             "volumetric_weight_divisor_factor": {
-              "type": "integer",
-              "default": 1
+              "type": "number",
+              "format": "float",
+              "default": 1.0
             },
             "created_at": {
               "$ref": "#/components/schemas/Timestamp"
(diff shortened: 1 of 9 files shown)


specification/schemas/contracts/Contract.json
```diff
@@ -19,6 +16,13 @@
               "type": "string",
               "example": "active"
             },
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            },
+            "volumetric_weight_divisor_factor": {
+              "type": "integer",
```
M4tini, line 23: This needs to be a float or number

specification/schemas/contracts/Contract.json
```diff
@@ -19,6 +16,14 @@
...
               "example": "active"
             },
+            "currency": {
+              "$ref": "#/components/schemas/Currency"
+            },
+            "volumetric_weight_divisor_factor": {
+              "type": "number",
+              "format": "float",
+              "default": 1.0
+            },
```
NickVries, line 26: Perhaps a short description could be useful here.

NickVries (changes requested):

NickVries (approved):

PR #1076, opened 2024-08-12
:sparkles: Added an include for hook on hook logs.

NickVries:
## Changes
- Added an include for hooks on hook logs api.

## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-6809


M4tini (approved):

NickVries:
Closing this because the proposed solution will not work as expected due to ES being involved. Replacing this with a new ticket in Jira to fix the issue in a different manner.

PR #1078, opened 2024-08-13
:fire: Remove unnecessary return items properties

yoerriwalstra:
Changes
- :fire: remove properties: sku, name, description, image_url, item_price, item_weight, features

Related issue
https://myparcelcombv.atlassian.net/browse/MP-6810

TODO
* [ ] merge together with [API changes](https://github.com/MyParcelCOM/api/pull/3375) to prevent pipeline from breaking


specification/schemas/returns/ReturnItem.json
```diff
@@ -3,69 +3,17 @@
   "additionalProperties": false,
   "required": [
     "id",
-    "quantity",
-    "description",
-    "item_price"
+    "quantity"
```
yoerriwalstra, line 6: Note to self: all these removals should only have effect on ReturnItems in a POST request. Will split this schema into GetReturnItem and PostReturnItem

yoan-myparcel (approved):

PR #1079, opened 2024-08-14
[api-spec] :sparkles: Return questions

zwaans:
https://myparcelcombv.atlassian.net/browse/MP-6792


specification/schemas/return-questions/ReturnQuestionDescription.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "object",
```
yoerriwalstra, line 2: Should we refactor the ReturnReasonManagementDescription into a general LocalizedDescription spec? Seems to me like we only have to give it a generic `description` and `example` and we can avoid adding a distinct spec for every resource with a translatable description

specification/paths/Returns-v1-questions.json
```diff
@@ -0,0 +1,161 @@
...
+  "get": {
+    "tags": [
+      "ReturnQuestions"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
+        ]
+      },
```
yoerriwalstra, line 11:
```suggestion
```

specification/paths/Returns-v1-questions-return_question_id.json
```diff
@@ -0,0 +1,136 @@
...
+  "get": {
+    "tags": [
+      "ReturnQuestions"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
+        ]
+      },
```
yoerriwalstra, line 19:
```suggestion
```

yoerriwalstra (approved):

PR #1080, opened 2024-08-14
✨  Introduced `/returns/v1/reports` endpoints

yoan-myparcel:


specification/schemas/return-reports/ReturnReport.json
```diff
@@ -0,0 +1,225 @@
...
+                "brokers": {
+                  "type": "array",
+                  "description": "Array of broker IDs to filter included returns by.",
+                  "items": {
+                    "type": "string",
+                    "format": "uuid",
+                    "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
+                    "example": "a294ee55-bc94-4890-b734-afd56c158f95"
+                  }
+                },
```
yoerriwalstra, line 115: I would re-use the `Uuid` schema and set `example` on the parent if you want to showcase a specific value in the preview JSON, but this is more of a preference, so I'll just leave it as a comment and you can decide what you want to do with it
```suggestion
                "brokers": {
                  "type": "array",
                  "description": "Array of broker IDs to filter included returns by.",
                  "items": {
                    "$ref": "#/components/schemas/Uuid"
                  },
                  "example": ["a294ee55-bc94-4890-b734-afd56c158f95"]
                },
```

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,194 @@
...
+      }
+    },
+    "return_statuses": {
+      "type": "array",
+      "description": "Comma separated string of return status codes to filter by. Only entries that correspond with one of the return status codes will be in the report.",
+      "items": {
+        "type": "string",
+        "example": "return-pending,return-approved"
+      }
+    },
```
yoerriwalstra, line 172: I suspect something is going wrong here (with both `return_reasons` and `return_statuses`). It makes sense to have a comma separated string of codes/statuses, _or_ have an array of strings. Since this is a POST body and we don't need to put everything in the URL, I'd opt for the array of strings. So, instead of
```
"items": {
  "example": "return-pending,return-approved" <<<---- this is an example of a single array element
}
```
we get
```
"items": {
  "type": "string"
},
"example": [
  "return-pending",
  "return-approved"
]
```
you should also update the description to reflect these changes

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,194 @@
...
+      "example": "exchange"
+    },
+    "return_methods": {
+      "type": "array",
+      "description": "Array of return method IDs to filter included returns by.",
+      "items": {
+        "type": "string",
+        "format": "uuid",
+        "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$"
+      },
```
yoerriwalstra, line 125:
```suggestion
      "items": {
        "$ref": "#/components/schemas/Uuid"
      },
```

specification/paths/Returns-v1-reports-count-report-entries.json
```diff
@@ -0,0 +1,73 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "responses": {
+      "201": {
```
yoerriwalstra, line 47: Do we return `201` even if we don't create a record in a table somewhere?
```suggestion
      "200": {
```

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+    },
+    "return_reasons": {
+      "type": "array",
+      "description": "Comma separated string of return reason codes to filter by. Only entries that correspond with one of the return reason codes will be in the report.",
+      "items": {
+        "type": "string"
+      },
+      "example": "product-damaged,product-is-defect"
```
yoerriwalstra, line 143:
```suggestion
      "example": [
        "product-damaged",
        "product-is-defect"
      ]
```

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,174 @@
...
+    },
+    "return_statuses": {
+      "type": "array",
+      "description": "Comma separated string of return status codes to filter by. Only entries that correspond with one of the return status codes will be in the report.",
+      "items": {
+        "type": "string"
+      },
+      "example": "return-pending,return-approved"
```
yoerriwalstra, line 151:
```suggestion
      "example": [
        "return-pending",
        "return-approved"
      ]
```

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,174 @@
...
+        "date_to": {
+          "type": "string",
+          "description": "Date string in YYYY-MM-DD format. Only returns paid at <= this date will be in the response.",
+          "example": "2024-03-01"
+        }
+      }
+    },
+    "return_reasons": {
+      "type": "array",
+      "description": "Comma separated string of return reason codes to filter by. Only entries that correspond with one of the return reason codes will be in the report.",
```
yoerriwalstra, line 139:
```suggestion
      "description": "Array of return reason codes to filter by. Only entries that correspond with one of the return reason codes will be in the report.",
```

specification/schemas/return-reports/ReturnReportFilters.json
```diff
@@ -0,0 +1,174 @@
...
+      "items": {
+        "type": "string"
+      },
+      "example": "product-damaged,product-is-defect"
+    },
+    "return_statuses": {
+      "type": "array",
+      "description": "Comma separated string of return status codes to filter by. Only entries that correspond with one of the return status codes will be in the report.",
```
yoerriwalstra, line 147:
```suggestion
      "description": "Array of return status codes to filter by. Only entries that correspond with one of the return status codes will be in the report.",
```

yoerriwalstra (approved):

PR #1083, opened 2024-08-19
:sparkles: Add endpoints to link and unlink question to shop return reason

yoerriwalstra:
**Related issue**
https://myparcelcombv.atlassian.net/browse/MP-6822

### DIFF
--- a/specification/paths/Returns-v1-shops-shop_id-reasons-return_reason_id-questions.json
+++ b/specification/paths/Returns-v1-shops-shop_id-reasons-return_reason_id-questions.json
@@ -34,11 +34,11 @@
               "data": {
                 "type": "object",
                 "required": [
-                  "return_question_ids"
+                  "return_questions"
                 ],
                 "additionalProperties": false,
                 "properties": {
-                  "return_question_ids": {
+                  "return_questions": {
                     "description": "Array of objects with a return question code and a boolean indicating if the question is required.",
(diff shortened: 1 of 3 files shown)


specification/paths/Returns-v1-shops-shop_id-reasons-return_reason_id-questions.json
```diff
@@ -0,0 +1,83 @@
...
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "type": "object",
+                "required": [
+                  "return_question_ids"
```
zwaans, line 37:
```suggestion
                  "return_questions"
```

I think this should be called questions since we aren't supplying id's

specification/paths/Returns-v1-shops-shop_id-reasons-return_reason_id-questions.json
```diff
@@ -0,0 +1,83 @@
...
+                ],
+                "additionalProperties": false,
+                "properties": {
+                  "return_question_ids": {
```
zwaans, line 41:
```suggestion
                  "return_questions": {
```

zwaans (changes requested):

zwaans (approved):

PR #1087, opened 2024-08-22
Link return questions to a shop return reason: update path and description

yoerriwalstra:


specification/schemas/return-reasons/ShopReturnReason.json
```diff
@@ -40,6 +40,29 @@
...
+            },
+            "questions": {
+              "type": "array",
+              "items": {
+                "required": [
+                  "return_question_code",
+                  "is_required"
+                ],
+                "properties": {
+                  "return_question_code": {
```
yoan-myparcel, line 52: Why not just `code`?

yoan-myparcel (changes requested):

yoan-myparcel (approved):

PR #1091, opened 2024-08-26
:sparkles: Introduced questions on return.item.reason

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-6799


specification/schemas/returns/PostReturnItem.json
```diff
@@ -59,6 +59,13 @@
             "null"
           ],
           "description": "Additional information regarding the reason for the return."
+        },
+        "questions": {
+          "type": "array",
+          "description": "The questions and answers related to this reason.",
```
zwaans, line 65:
```suggestion
```

PR #1109, opened 2024-09-25
✨ Contract templates

M4tini:
https://myparcelcombv.atlassian.net/browse/MP-6893
---
- Make get-all.
- Make post.
- I decided to also make get-one and delete.

### DIFF
--- a/specification/paths/Contracts-v1-templates.json
+++ b/specification/paths/Contracts-v1-templates.json
@@ -1,0 +1,177 @@
+{
+  "get": {
+    "tags": [
+      "ContractTemplates"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "contract_templates.manage"
+        ]
+      }
+    ],
+    "summary": "Get all contract templates.",
+    "description": "This endpoint retrieves all contract templates available to you.",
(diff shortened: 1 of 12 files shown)


specification/paths/Contracts-v1-templates.json
```diff
@@ -0,0 +1,177 @@
...
+                  },
+                  {
+                    "required": [
+                      "attributes",
+                      "relationships"
+                    ],
+                    "properties": {
+                      "id": {
+                        "readOnly": true
+                      },
```
yoerriwalstra, line 126: this feels more like something that would go on a PUT/PATCH request

yoerriwalstra (approved):

specification/paths/Contracts-v1-templates.json
M4tini: @yoerriwalstra The resource schema we put in the `allOf` has a `type` and `id` and we have to define this readOnly to disallow posting the `id`.

It would be nice to find a way to avoid defining this readOnly, but if we don't include the `id` in the resource schema, then we would have to define it in a different way. Might be worth looking into.

PR #1117, opened 2024-10-16
:sparkles: introduce ReturnMethod title and description

yoerriwalstra:
**Changes**
- :sparkles: introduce ReturnMethod title and description
- :fire: remove name

**Related issue**
https://myparcelcombv.atlassian.net/browse/MP-6935

### DIFF
--- a/specification/schemas/return-methods/ReturnMethod.json
+++ b/specification/schemas/return-methods/ReturnMethod.json
@@ -20,9 +20,15 @@
             "default_weight": {
               "$ref": "#/components/schemas/Weight"
             },
-            "name": {
+            "title": {
               "type": "string",
-              "description": "Required if the return_type attribute is `shipment`."
+              "description": "Title to display to the consumer in the return portal.",
+              "example": "Super simple return method"
+            },
+            "description": {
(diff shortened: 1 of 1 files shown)


specification/schemas/return-methods/ReturnMethod.json
```diff
@@ -20,9 +20,17 @@
             "default_weight": {
               "$ref": "#/components/schemas/Weight"
             },
-            "name": {
+            "title": {
```
M4tini, line 23: Shall we remove the `"additionalProperties": false` above to make old requests not fail on the spec?

yoerriwalstra: Good call, I thought about this after I made the PR but forgot to bring up the question. I wondered though, since we're the only consumers of this API, if we could get away with this (breaking) change and keep the `additionalProperties` property?

M4tini: If all our code is updated within the same release, then that will be fine 👌🏻

M4tini (approved):

PR #1133, opened 2024-11-19
👌🏻  Added optional fields for return emails customer service phone and email

yoan-myparcel:


specification/schemas/shops/Shop.json
```diff
@@ -125,6 +125,18 @@
...
+                            "string",
+                            "null"
+                          ],
+                          "description": "Phone number for customer service",
+                          "example": "+31 20 123 4567",
+                          "maxLength": 20
+                        },
+                        "customer_service_email": {
+                          "$ref": "#/components/schemas/Email"
+                        },
```
yoerriwalstra, line 139: considering this field is nullable and to make it in line with the rest of the email content schema definitions, I think we should make this either schemas/Email, or null
```suggestion
                        "customer_service_email": {
                          "oneOf": [
                            {
                              "$ref": "#/components/schemas/Email"
                            },
                            {
                              "type": "null"
                            }
                          ]
                        },
```

yoerriwalstra (approved):

PR #1138, opened 2024-11-22
👌🏻  Added carrier and service fields in return methods meta for easy access

yoan-myparcel:
# What's Changed
Some basic carrier and service fields like name, code and logo URL are always needed when the resolve return methods endpoint will be called. Therefore, it is not necessary to have them as relationship includes

### DIFF
--- a/specification/paths/Returns-v1-return-methods.json
+++ b/specification/paths/Returns-v1-return-methods.json
@@ -17,14 +17,13 @@
         "$ref": "#/components/parameters/query-filter-shop"
       },
       {
-        "allOf": [
-          {
-            "$ref": "#/components/parameters/query-filter-weight"
-          },
-          {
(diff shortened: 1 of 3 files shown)


specification/paths/Returns-v1-return-methods.json
```diff
@@ -17,14 +17,13 @@
...
-            "deprecated": true
-          }
-        ]
+        "name": "filter[weight]",
+        "in": "query",
+        "deprecated": true,
+        "description": "Weight value in grams to filter by.",
+        "schema": {
+          "type": "number"
+        }
```
yoan-myparcel, line 26: I had to define it like that because merging with deprecated flag was making the entire filter disappear from the rendered spec.

yoerriwalstra (approved):

PR #1147, opened 2025-01-06
MP-7072 Last login user attribute

daniel-myparcel:
Changes:
- Added a `last_login` to the User schema
- Added the `filter[last_login_before]` to enable filtering users that have not been logged in for a while
- Exposed `last_login` in the UserResponse schema

Ticket: https://myparcelcombv.atlassian.net/browse/MP-7072

### DIFF
--- a/specification/paths/Users.json
+++ b/specification/paths/Users.json
@@ -26,11 +26,20 @@
         }
       },
       {
+        "$ref": "#/components/parameters/query-filter-created_at-date_from"
+      },
+      {
+        "$ref": "#/components/parameters/query-filter-created_at-date_to"
+      },
+      {
         "name": "filter[last_login_before]",
         "in": "query",
-        "description": "Retrieve users who last logged in before the specified time.",
+        "description": "Date string in ISO 8601 format or unix timestamp. Only resources created at >= this date will be in the response.",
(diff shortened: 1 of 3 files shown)


specification/paths/Users.json
```diff
@@ -24,6 +24,14 @@
         "schema": {
           "type": "string"
         }
+      },
+      {
+        "name": "filter[last_login_before]",
+        "in": "query",
+        "description": "Retrieve users who last logged in before the specified time.",
+        "schema": {
+          "$ref": "#/components/schemas/Timestamp"
```
yoerriwalstra, line 33: Looking at other date(time) query params, we usually accept a [date string](https://github.com/MyParcelCOM/api-specification/blob/develop/specification/parameters/query-filter-date_from.json), or [date strings and timestamps](https://github.com/MyParcelCOM/api-specification/blob/develop/specification/parameters/query-filter-created_at-date_from.json). My guess is only a date string is fine here, but might want to confirm with one of the other devs

M4tini: This attribute is read-only and our API output is always a unix timestamp in UTC timezone.

yoerriwalstra (approved):

PR #1150, opened 2025-01-08
:sparkles: Add `address` property to ReturnMethod and Return schemas

yoerriwalstra:
**Related issue**
https://myparcelcombv.atlassian.net/browse/MP-7080

**TODO**
* [ ] merge together with related API PR (under construction) to prevent pipelines from breaking over missing required property in the response

### DIFF
--- a/specification/paths/Returns-v1-returns.json
+++ b/specification/paths/Returns-v1-returns.json
@@ -178,16 +178,7 @@
                           "order_reference",
                           "items",
                           "consumer_address"
-                        ],
-                        "properties": {
-                          "consumer_address": {
-                            "required": [
-                              "street_1",
-                              "city",
-                              "country_code"
-                            ]
-                          }
-                        }
+                        ]
                       },
                       "relationships": {
                         "required": [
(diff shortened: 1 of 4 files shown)


specification/paths/Returns-v1-returns.json
```diff
@@ -178,16 +178,7 @@
...
-                        ],
-                        "properties": {
-                          "consumer_address": {
-                            "required": [
-                              "street_1",
-                              "city",
-                              "country_code"
-                            ]
-                          }
-                        }
```
M4tini, line 190: Why is this removed? If there is a `consumer_address` we still require it to have these three attributes, right?

M4tini (changes requested):

specification/paths/Returns-v1-returns.json
M4tini: Ah, these are already set in the `Return` model. We normally define requirements in the `...Response` model and for post requests in the endpoint definition (otherwise a patch request cannot update only 1 attribute). But since we do not have patch requests for returns, it's fine like this.

M4tini (approved):

specification/paths/Returns-v1-returns.json
yoerriwalstra: oops, thought I left a comment on this myself, but you figured it out :+1:

PR #1162, opened 2025-01-24
MP-7122 Add countries from attribute to contract template schema

NickVries:
## Changes
- ✨ Added `countries_from` attribute to ContractTemplate schema.
- ✨ Added `filter[countries_from]` query param to the GET endpoint for contract templates

## Related Issues
- https://myparcelcombv.atlassian.net/browse/MP-7122

## Notes 
I implore the reviewer to come up with an alternative name for this attribute if they can think of one.

### DIFF
--- a/specification/paths/Contracts-v1-templates.json
+++ b/specification/paths/Contracts-v1-templates.json
@@ -26,6 +26,14 @@
       },
       {
         "$ref": "#/components/parameters/query-filter-carrier"
+      },
+      {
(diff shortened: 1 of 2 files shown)


M4tini (approved):
On a service we call it `regions_from` so I think `countries_from` is in line with that implementation. Also having in the back of my mind that we're going to use zones in the (hopefully not too far) future and we are the only consumers of this endpoint.

PR #1171, opened 2025-03-21
✨  Added first_mile_carrier attribute to Shipment schema

yoan-myparcel:
# What's Changed
Added a new Shipment schema field for First mile carrier name

### DIFF
--- a/specification/schemas/shipments/Shipment.json
+++ b/specification/schemas/shipments/Shipment.json
@@ -155,6 +155,19 @@
                 }
               }
             },
+            "first_mile_carrier": {
+              "readOnly": true,
+              "type": "object",
+              "required": [
+                "name"
+              ],
+              "properties": {
+                "name": {
+                  "type": "string",
+                  "description": "The name of the first mile carrier"
+                }
+              }
+            },
             "sync_active": {
               "readOnly": true,
               "type": "boolean",


specification/schemas/shipments/Shipment.json
```diff
@@ -155,6 +155,19 @@
                 }
               }
             },
+            "first_mile_carrier": {
+              "readOnly": true,
```
yoerriwalstra, line 159: According to [the ticket in Jira](https://myparcelcombv.atlassian.net/browse/MP-6903), the `first_mile_carrier` is exposed in the response from the _carrier specification_, does it have to go here as well? Maybe it's required to pass our schema validation? It does mean that `first_mile_carrier` shows up in the response specification now, while the ticket mentions that "We do not need to expose this field through the API responses".

yoerriwalstra: Update, I see now in the API PR that we _are_ exposing the new property through the transformer, so I'll take this as an indication that it proved to be necessary for the feature, correct?

yoan-myparcel: You are right, we do not need it in the api response. I will clean it up!

yoan-myparcel:
I am closing this PR as indeed the first mile carrier does not need to be exposed.

PR #1183, opened 2025-04-18
:lipstick: [api-spec] Added carrier code as a required field to the delivery dates endpoint

zwaans:
# https://myparcelcombv.atlassian.net/browse/MP-7273

### DIFF
--- a/specification/paths/Shipping-v2-available-delivery-dates.json
+++ b/specification/paths/Shipping-v2-available-delivery-dates.json
@@ -28,6 +28,7 @@
                 "type": "object",
                 "required": [
                   "service_code",
+                  "carrier_code",
                   "address",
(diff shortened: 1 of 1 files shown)


specification/paths/Shipping-v2-available-delivery-dates.json
```diff
@@ -38,6 +39,12 @@
                     "description": "The service code for which to fetch the available delivery date time windows.",
                     "example": "carrier-parcel-next-day"
                   },
+                  "carrier_code": {
```
M4tini, line 42: Let's position `carrier_code` as the first attribute, because the `service_code` depends on this value

M4tini (changes requested):
+ make from master

PR #1188, opened 2025-04-28
✨  Added endpoint that allows an order reference to be reissued as a return

yoan-myparcel:
# What's Changed
- Added a simple endpoint that allows an order reference to be reissued on a shop

### DIFF
--- a/specification/paths/Returns-v1-returns-allow-reissue.json
+++ b/specification/paths/Returns-v1-returns-allow-reissue.json
@@ -1,0 +1,68 @@
+{
+  "post": {
+    "tags": [
+      "Returns"
+    ],
+    "security": [
+      {
+        "OAuth2": [
+          "returns.manage"
+        ]
+      }
+    ],
+    "summary": "Allow an order reference to be reissued as a return.",
+    "description": "This endpoint allows a return to be reissued for an existing order reference.",
+    "requestBody": {
+      "description": "The shop ID and the order reference.",
+      "required": true,
+      "content": {
+        "application/vnd.api+json": {
+          "schema": {
+            "type": "object",
+            "required": [
+              "data"
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "type": "object",
+                "required": [
+                  "shop_id",
+                  "order_reference"
+                ],
+                "additionalProperties": false,
+                "properties": {
+                  "shop_id": {
+                    "type": "string",
+                    "format": "uuid",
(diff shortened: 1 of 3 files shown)


specification/paths/Returns-v1-returns-allow-reissue.json
```diff
@@ -0,0 +1,68 @@
...
+                  "order_reference"
+                ],
+                "additionalProperties": false,
+                "properties": {
+                  "shop_id": {
+                    "type": "string",
+                    "format": "uuid",
+                    "pattern": "^[a-f\\d]{8}(-[a-f\\d]{4}){3}-[a-f\\d]{12}$",
+                    "example": "4067e35b-09cc-42b9-8553-6eaf1788df35"
+                  },
```
yoerriwalstra, line 40: We have a `Uuid` schema to reference
```suggestion
                  "shop_id": {
                    "$ref": "#/components/schemas/Uuid"
                  },
```
but it does not include an example, so to add one, I believe you would have to make an `allOf` construction, which is a bit verbose...
```suggestion
                  "shop_id": {
                    "allOf": [
                      {
                        "$ref": "#/components/schemas/Uuid"
                      },
                      {
                        "example": "4067e35b-09cc-42b9-8553-6eaf1788df35"
                      }
                    ]
                  },
```
I'll leave it up to you if you want to make any changes

specification/paths/Returns-v1-returns-allow-reissue.json
```diff
@@ -0,0 +1,68 @@
...
+        "description": "The order reference can be issued again one more time."
+      },
+      "404": {
+        "description": "Shop not found."
+      },
+      "422": {
+        "description": "There is no existing return with that order reference - no need to allow reissue."
+      },
+      "409": {
+        "description": "There ia pending unclaimed reissue for this order reference and shop."
```
yoerriwalstra, line 64:
```suggestion
        "description": "There is a pending unclaimed reissue for this order reference and shop."
```

specification/schemas/statuses/Status.json
```diff
@@ -29,6 +29,7 @@
                 "return-payment-failed",
                 "return-pending",
                 "return-rejected",
+                "return-rejected-for-reissue",
```
yoerriwalstra, line 32: Maybe it's just me, but I don't really understand why we would say a return is "rejected" for reissue. Does reissue-ing a return imply rejecting the previous/original one? Wouldn't it be simpler to only use `return-reissued`?

yoerriwalstra (approved):

PR #1194, opened 2025-05-15
♻️ Remove `country` filter and add `broker` filter to the system-messages endpoint

M4tini:
- We do not use the `country_code` attribute or the `country` filter when working with system messages.
- We do want to start using the `broker` filter when scoping the backoffice to a single broker.
- I also added a missing entry to the `parameters.json`.

### DIFF
--- a/specification/paths/SystemMessages.json
+++ b/specification/paths/SystemMessages.json
@@ -14,12 +14,7 @@
     "description": "This endpoint retrieves system messages. These are informative messages about our system or a specific type of resource.",
     "parameters": [
       {
-        "name": "filter[country]",
-        "in": "query",
-        "description": "Country code of system messages to filter by. Using this filter will also include system-wide messages.",
-        "schema": {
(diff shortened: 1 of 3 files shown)


yoerriwalstra (approved):

specification/paths/SystemMessages.json
```diff
@@ -14,12 +14,7 @@
     "description": "This endpoint retrieves system messages. These are informative messages about our system or a specific type of resource.",
     "parameters": [
```
yoerriwalstra, line 15: looking at the API PR, a broker can now be included, so we should add a query param for it
```
{
  "name": "include",
  "in": "query",
  "description": "Comma separated string of the relationship names you want to include the data of.<br>The relationships that can be included are:<ul><li>`broker`</li>",
  "schema": {
    "type": "string"
  }
},
```

yoerriwalstra: https://github.com/MyParcelCOM/api-specification/pull/1195

PR #1202, opened 2025-06-11
Added return portal URL to shop resource

daniel-myparcel:
https://myparcelcombv.atlassian.net/browse/MP-7305

### DIFF
--- a/specification/schemas/shops/ShopResponse.json
+++ b/specification/schemas/shops/ShopResponse.json
@@ -35,6 +35,14 @@
                 "country_code"
               ]
             },
+            "return_portal_url": {
+              "type": [
+                "string",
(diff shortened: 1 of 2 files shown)


specification/schemas/shops/ShopResponse.json
```diff
@@ -17,6 +17,7 @@
             "return_address",
             "sender_address",
             "created_at",
+            "return_portal_url",
```
daniel-myparcel, line 20: @M4tini Does it make sense to keep `return_portal_url` here?

M4tini: Yes, this will make sure it's shown as a required attribute in the response, meaning we always output it

PR #1204, opened 2025-06-17
Bug + incorrect type fix

daniel-myparcel:
The `Shop` should have not had the `additionalProperties=false`. It made the schema fail as now we have a required property `return_portal_url` which is part of the `ShopResponse`.

I used this as an opportunity to fix another inaccuracy which is that `return_portal_url` will always resolve to a valid string making `null` an invalid type. Fixed as well

NOTE: we went with @NickVries suggestions of utilizing `readOnly` [here](https://github.com/MyParcelCOM/api-specification/pull/1204#discussion_r2151872154)

### DIFF
--- a/specification/schemas/shops/Shop.json
+++ b/specification/schemas/shops/Shop.json
@@ -223,7 +223,7 @@
               "type": "string",
               "readOnly": true,
               "example": "https://my-return-portal-domain.com",
-              "description": "The domain used as the return portal for this shop. If not set, defaults to the return portal domain of the broker followed by the shop's name. E.g. `https://return.myparcel.com/my-shop`"
+              "description": "The URL used as the return portal for this shop. If not set, defaults to the return portal domain of the broker followed by the shop's name. E.g. `https://return.myparcel.com/my-shop`"
             },
             "return_portal_locales": {
               "type": "array",
--- a/specification/schemas/shops/Shop.json
+++ b/specification/schemas/shops/Shop.json
@@ -7,6 +7,7 @@
       "properties": {
         "attributes": {
           "type": "object",
+          "additionalProperties": false,
           "properties": {
             "name": {
               "type": "string",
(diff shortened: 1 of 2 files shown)


specification/schemas/shops/Shop.json
```diff
@@ -7,7 +7,6 @@
       "properties": {
         "attributes": {
           "type": "object",
-          "additionalProperties": false,
```
NickVries, line 10: Let's leave this additionalProperties: false here, since it protects from sending attributes that are not defined on the schema.

Instead, we can define the `return_portal_url` attribute in the `Shop` schema and make it readOnly (by adding `"readOnly": true` to the property). 

Don't forget to add the `return_portal_url` attribute in the `required` array of the `ShopResponse` schema ;)

specification/schemas/shops/ShopResponse.json
```diff
@@ -35,11 +35,6 @@
                 "country_code"
               ]
             },
-            "return_portal_url": {
-              "type": "string",
-              "example": "https://my-return-portal-domain.com",
-              "description": "The domain used as the return portal for this shop. If not set, defaults to the return portal domain of the broker followed by the shop's name. E.g. `https://return.myparcel.com/my-shop`"
-            },
```
NickVries, line 45: Even though it's removed here, we still need to add it to the list of `required` attributes in the `ShopResponse`.

NickVries (changes requested):

specification/schemas/shops/ShopResponse.json
daniel-myparcel: It is already there

daniel-myparcel: From the original PR

specification/schemas/shops/Shop.json
```diff
@@ -223,7 +223,7 @@
               "type": "string",
               "readOnly": true,
               "example": "https://my-return-portal-domain.com",
-              "description": "The URL used as the return portal for this shop. If not set, defaults to the return portal domain of the broker followed by the shop's name. E.g. `https://return.myparcel.com/my-shop`"
+              "description": "The URL used for the return portal for this shop. If not set, defaults to the return portal domain of the broker followed by the shop's name. E.g. `https://return.myparcel.com/my-shop`"
```
NickVries, line 226: Let's also put the example that's in the description as the example for this field!

NickVries (changes requested):

NickVries (approved):

PR #1231, opened 2025-09-26
locale enum for invitation request

daniel-myparcel:
https://myparcelcombv.atlassian.net/browse/MP-7539

### DIFF
--- a/specification/schemas/invitations/InvitationRequest.json
+++ b/specification/schemas/invitations/InvitationRequest.json
@@ -47,6 +47,11 @@
                   "returns"
                 ]
               }
+            },
+            "locale": {
+              "$ref": "#/components/schemas/LocaleEnum",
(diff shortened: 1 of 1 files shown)


specification/schemas/invitations/InvitationRequest.json
yoerriwalstra, line 1: it's not very common that we have separate `...Request` classes in our specification. Most of the time, we would have a path specification, such as `Onboarding-v1-invitations.json`, which would reference a resource, such as `Invitation.json`, in its request body specification. I see we have both a `InvitationRequest` and an `Invitation` specification, with significant overlap. 

Can you investigate the origin of this differentiation and if we can perhaps remove the `...Request` class entirely?

If that is not possible/feasible, I would at least expect the `Invitation` specification to also receive this new property

PR #1291, opened 2026-05-19
Files endpoint for a return

daniel-myparcel:
This is a new endpoint to offer labels as download in the return portal.

https://myparcelcombv.atlassian.net/browse/MP-7875

### DIFF
--- a/specification/paths/Returns-v1-returns-return_id-files.json
+++ b/specification/paths/Returns-v1-returns-return_id-files.json
@@ -15,8 +15,8 @@
         ]
       }
     ],
-    "summary": "Get all files for a return.",
-    "description": "Retrieves the files associated with the shipment belonging to the given return.",
+    "summary": "Get all the files from a return.",
+    "description": "This endpoint retrieves files belonging to the shipment associated with the return.",
     "parameters": [
       {
         "name": "filter[document_type]",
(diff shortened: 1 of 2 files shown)


specification/paths/Returns-v1-returns-return_id-files.json
```diff
@@ -0,0 +1,58 @@
...
+        "OAuth2": [
+          "returns.manage"
+        ]
+      }
+    ],
+    "summary": "Get all the files from a return.",
+    "description": "This endpoint retrieves files belonging to the shipment associated with the return.",
+    "parameters": [
+      {
+        "name": "filter[document_type]",
```
zwaans, line 22: The ticket describes that this should be handled via a header?

zwaans (changes requested):

specification/paths/Returns-v1-returns-return_id-files.json
daniel-myparcel: From looking at other requests in the spec I think this is pretty much the way we filter request data. Maybe its a mistake in the ticket description?

zwaans: Hmm I thought I responded to this. We did this because the files endpoint also handles it that way https://api-specification.myparcel.com/#tag/Files

zwaans (approved):

PR #1299, opened 2026-07-14
Add `sender_tracking_url` to shipment

daniel-myparcel:
https://myparcelcombv.atlassian.net/browse/MP-8319

### DIFF
--- a/specification/schemas/shipments/Shipment.json
+++ b/specification/schemas/shipments/Shipment.json
@@ -128,6 +128,7 @@
               "description": "URL used by the recipient to view tracking status updates."
             },
             "sender_tracking_url": {
+              "readOnly": true,
               "oneOf": [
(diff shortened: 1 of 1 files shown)


specification/schemas/shipments/Shipment.json
```diff
@@ -123,8 +123,21 @@
             "tracking_code": {
               "$ref": "#/components/schemas/TrackingCode"
             },
-            "tracking_url": {
-              "$ref": "#/components/schemas/TrackingUrl"
+            "recipient_tracking_url": {
```
daniel-myparcel: Im going to add `tracking_url` back and mark it as deprecated since its safer to keep the old functionality intact while I still think we should rename this field

yoerriwalstra (approved):

PR #1302, opened 2026-07-22
Accept lowercase country codes in JSON Schema

daniel-myparcel:
## What changed
- :sparkles: Introduced `CountryCodeUserInput` for cases country codes are passed to us by a user
- :recycle: `CountryCode` made into ENUM for better use of JSON Schema conventions
- :recycle: Update `path-country_code` to use `CountryCodeUserInput`

NOTE: The reason I introduced the additional `CountryCodeUserInput` is because when we return a country code it will always be uppercase so changing that schema would be misleading.

[Ticket](https://myparcelcombv.atlassian.net/browse/MP-8335)


specification/schemas/addresses/CountryCode.json
```diff
@@ -1,6 +1,32 @@
...
+    "RU", "RW", "SA", "SB", "SC", "SD", "SE", "SG", "SH", "SI",
+    "SJ", "SK", "SL", "SM", "SN", "SO", "SR", "SS", "ST", "SV",
+    "SX", "SY", "SZ", "TC", "TD", "TF", "TG", "TH", "TJ", "TK",
+    "TL", "TM", "TN", "TO", "TR", "TT", "TV", "TW", "TZ", "UA",
+    "UG", "UM", "US", "UY", "UZ", "VA", "VC", "VE", "VG", "VI",
+    "VN", "VU", "WF", "WS", "XK", "YE", "YT", "ZA", "ZM", "ZW"
+  ],
   "example": "GB",
-  "description": "Country code in ISO 3166-1 alpha-2 format."
+  "description": "Country code in ISO 3166-1 alpha-2 format (uppercase or lowercase)."
```
M4tini: This one is always uppercase.
```suggestion
  "description": "Country code in ISO 3166-1 alpha-2 format."
```

specification/schemas/addresses/CountryCodeUserInput.json
```diff
@@ -0,0 +1,57 @@
...
+    "mr", "ms", "mt", "mu", "mv", "mw", "mx", "my", "mz", "na",
+    "nc", "ne", "nf", "ng", "ni", "nl", "no", "np", "nr", "nu",
+    "nz", "om", "pa", "pe", "pf", "pg", "ph", "pk", "pl", "pm",
+    "pn", "pr", "ps", "pt", "pw", "py", "qa", "re", "ro", "rs",
+    "ru", "rw", "sa", "sb", "sc", "sd", "se", "sg", "sh", "si",
+    "sj", "sk", "sl", "sm", "sn", "so", "sr", "ss", "st", "sv",
+    "sx", "sy", "sz", "tc", "td", "tf", "tg", "th", "tj", "tk",
+    "tl", "tm", "tn", "to", "tr", "tt", "tv", "tw", "tz", "ua",
+    "ug", "um", "us", "uy", "uz", "va", "vc", "ve", "vg", "vi",
+    "vn", "vu", "wf", "ws", "xk", "ye", "yt", "za", "zm", "zw"
```
M4tini, line 53: I get the heebie jeebies from this enum and am wondering why we're doing all this work in the specification.
The bug report is caused by the issue that our API forwards the path parameter without making it uppercase.
The minimal fix for this would be a `strtoupper()` in the API and we're done, or am I overlooking something?

It's nice to make the regex an enum, but I think we should leave that as the only change in this PR.

M4tini (changes requested):

specification/schemas/addresses/CountryCodeUserInput.json
daniel-myparcel: That's a good question. The ticket title says the issue is `[api-specification] lowercase country code accepted` which insinuates the bug is in the spec (since we also use it to validate the request body).

Now that I read the ticket again getting an empty response as the ticket mentions menas it does infact pass validation and so this is not the bug.

daniel-myparcel: I think that even if this doesn't fix an actual bug in the flow we might want to keep this for semantic purposes.
The Spec does accepts lowercase country codes so having it reflected in the Spec is arguably logical

daniel-myparcel: https://github.com/MyParcelCOM/api/pull/4313

